PR #41, opened 2020-11-24
Refactor command-line handling

djmitche:


djmitche:
I'm not totally happy with the use of macros here, but it's good enough to stop futzing with it.  I started writing the macros because I wanted to "hide" the Any/downcast_ref stuff, but in the end I probably made things more confusing rather than less.

PR #96, opened 2020-12-17
subcommands: prepend and append to description

### LINKED ISSUE #88: Add "prepend" and "append" commands
These just add new text to the beginning or end of the task description.

srfraser:
This currently uses format!() which may not be the best option. Fixes #88

I'm unsure how to set up an existing test task to ensure the test does actually append and prepend as desired.


djmitche (approved):
Looks good!  I think `format!` is the right tool here.

PR #98, opened 2020-12-19
Switch to a command-line API closer to TaskWarrior

djmitche:
* Use a parser (rather than clap) to process the command line
* Outline some generic support for filtering, reporting, modifying, etc.
* Break argument parsing strictly from invocation, to allow independent testing

I'm still not happy with some of the abstractions like Modification and Filter, but those aren't a public API so there's always time for more refactoring.


djmitche:
@srfraser do you want to have a look?  I did include `append` and `prepend` as forms of Modification :)

djmitche:
The idea with the help support is that eventually it can be used to generate docs (the markdown stuff) and maybe a manpage.

djmitche:
I'm going to get this merged, since it blocks further work on the CLI, but I'm still happy to hear comments :)

Issue #11, opened 2020-11-21
Support filtering by tags


djmitche:
Similar syntax to TaskWarrior for adding, removing, and filtering by tags.

srfraser:
One thing that would be helpful for me is some streamlining around projects or tags.

My first thought is something a config option that would automatically add a tag. Perhaps if I have a global config option set my current working directory (and its parents up to `$HOME` are scanned for `.git` `.hg` and equivalent directories, allowing us to work out an automatic tag to add.

A task added in a clone of this repo could be tagged with `taskchampion` based on the directory or perhaps a remote, or the result of some external script we've been told to call (which lets people specify branches or ticket/bug numbers as they want)

djmitche:
I think we'd want to find another level integration, different from "run these two commands and if they don't produce an empty string .."

Let's take that part to a new issue.  I've moved the parent-directories thing to #79.

djmitche:
I'm going to break out *modifying* tags from filtering on them, as they are different operations and both pretty complex.

This will be the first real "filtering" operation.  I want to get to a point where we can write arbitrary filter expressions using a pretty "normal" language.  One that's not necessarily compatible with TW.  But at the top level, there's an "intuitive" format that TW captures pretty well, but is maybe a bit hard to implement.  For example:

```
task 1 2 abcd1234 <subcommand> --> task.id == 1 or task.id == 2 or task.uuid.starts_with('abcd2134')
task +next +food <subcommand> --> task.tags.has('next') and task.tags.has('food')
```

A few observations:
 * this "intuitive" filtering happens at the level of multiple command-line arguments
 * in some cases there's an implicit "or" between arguments (such as (UU)IDs)
 * in other cases there's an implicit "and" between arguments (such as tags).  I suspect this is the case for anything other than IDs?

So I think I want to build some fairly general support for filtering that constructs boolean trees, but accounts for this intuitiveness at the top level.

djmitche:
Maybe another way to look at this is that a filter has a universe and a conjunction of filter terms -- so "tasks from UNIVERSE matching COND1 AND COND2 AND ..".  If an id list is provided, that's the universe.  If no list is provided, the default universe is all tasks, but some subcommands or filter conditions may allow us to reduce that to just pending tasks (saving a lot of time in scanning the DB).

PR #107, opened 2020-12-21
Implement tags in the Replica API and the CLI

djmitche:
```
task add foo +bar
task 1 modify -bar +bing
task 1 info
```


djmitche:
Note that this doesn't support filtering by tags -- that comes in #11.

Issue #148, opened 2021-03-16
Switch to `anyhow` and `thiserror`


djmitche:
`failure` is deprecated.  `anyhow` is very similar!  And `thiserror` will be useful in the `taskchampion` crate to provide a rich error API to users.

PR #156, opened 2021-03-25
Replace 'failure' with anyhow + thiserror

dbr:
As per #148 - pretty much drop-in replacement so just a bunch of search-and-replace

I've sided on leaving anyhow fully-qualified, e.g `-> anyhow::Result<()>` etc instead of `use anyhow::Result; ... -> Result` - with the exception of `taskchampion/src/storage/mod.rs` where there it cluttered up the trait definition.

Haven't looked into using `thiserror` more extensively to give more informative errors from the core library

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -295,6 +295,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "anyhow"
+version = "1.0.39"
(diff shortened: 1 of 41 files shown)


djmitche (approved):
Looks great!

PR #176, opened 2021-03-28
fix some clippy stuff

djmitche:
Unnecessary wraps are fixed in one place and ignored in another (where they're required for a callback).

Capitalization fixed (note that this is an API-breaking change for KVStorage, but we're getting rid of that in favor of SQLite anyway)


dbr:
Looks like there's one other caught by the new linter, `DBError(String),`

Other than that, looks good to me

djmitche:
I hate that clippy caches and does not repeat its warnings

djmitche:
yay, clippy passed, finally :)

PR #206, opened 2021-04-22
SQLite backing store

dbr:
Very WIP - still to implement the methods regarding the working set, operations, and base version

For first pass I'm just using SQLite table as dumb key-value store:
- there's a tasks column with `uuid` and `data`
- `data` is a `serde_json` dump of the task data

Reason for using `serde_json` is mostly that it was already there, and JSON is reasonably simple/fast. Most probably not as fast as `Msgpack` which was used with LMDB, but quite possibly adequate?

There's no reason this couldn't be stored in a more SQL'y manner, with:

- tasks column with just UUID column
- task data table with `task_id`, `key` and `value`

However this feels like it might be a little excessive, and may end up being slower since the task-data table would end up being much larger

Other misc thoughts:

- The UUID is using "rusqlite" built-in handling, and stores the UUID a BLOB. It might make the database easier to work with manually if the UUID were stored as a string (via `Uuid::to_string`), but not sure this is really a concern.
- I'm using rusqlite as I've used it before, and it exposes a `Transaction` type which made the implementation simpler. The [`sqlite` project](https://docs.rs/sqlite/0.26.0/sqlite/index.html) does not. Both have been around for >5 years and still 0.x :cry:


djmitche (commented):
This is looking good!  People shouldn't be accessing the DB "manually", so I don't think there's any need to make that easier.  JSON is fine, but storing the data with msgpack would be A-OK too :)

(The reason being, the relationship between the tasks table and the operations table is rather precise, and Bad Things will happen if those get out of sync)

rusqlite seems like a good map to the needs of this app, particularly with its Transaction type.

dbr:
I've implemented the last of the methods and this seems to be functioning

Still to do:

1. Need some better (i.e non-unwrap) error handling for a few of the deserialization calls (think I just need to use rusqlite's serde feature properly)
2. Quick check over changes to find any obvious errors and see if I can neaten anything up

I'm also pretty tempted to store the UUID's as strings - for debugging purposes, from the sqlite3 interface it's pretty awkward with the current stored-as-blob approach:

```
sqlite> select * from tasks;
�n�KI����)o�|{"modified":"1619619386","description":"yep","status":"P"}
}{�:�gBt�S+���Ɯ|{"description":"(no description)","modified":"1619619389","status":"P"}
```

Then on to the sync-server storage

djmitche:
Yeah, that is pretty gross.  I don't think there's any real downside to storing strings.  Performance isn't a huge issue here.

dbr:
@djmitche The `trait Storage` requires `Send + Sync`, but.. rusqlite's Transaction type [is neither](https://docs.rs/rusqlite/0.25.1/rusqlite/struct.Transaction.html#impl-Send)

SQLite can be fully threadsafe but rusqlite isn't - [discussed here](https://github.com/rusqlite/rusqlite/issues/188), and also [here](https://github.com/rusqlite/rusqlite/issues/697), neither has too satisfying an outcome.

Any thoughts on the best way to implement this?

djmitche:
Ah, this has been a problem with Python as well.  Perhaps the storage should run in a dedicated thread, and the operations be performed via mpsc channels?  It might be easiest to just shuttle query text over the channel to the sqlite thread, with responses returned to the caller on dedicated channels.  The callers would then send the query text and `await` a message on the response channel.

dbr:
I've got a prototype of how the SQLite-in-thread-plus-async thing could work. The few difficulties I encountered were:

1. Me not having much experience with Rust's async (it seems quite easy to misuse, accidentally dropping futures without awaiting them properly makes for weird behaviour and no warnings etc)
2. A mpsc queue for sending query strings over is easy (enum with the various operations that are needed, containing a oneshot queue for returning the value) - that part is easy. However the rusqlite traits for query parameters also aren't `Send`, and so simplest solution is to probably wrap the parameters and return-types in our own simple type `enum ... { String(...), Uuid(...), Integer(...)}` or something). Doable, just extra work

This briefly lead me to consider just using sqlx, but https://github.com/deltachat/deltachat-core-rust/issues/2378 somewhat strongly put me of that idea

r2d2-sqlite seems interesting, however: https://docs.rs/r2d2_sqlite/0.18.0/r2d2_sqlite/index.html - a connection pool around rusqlite where the handle is send+sync. Not 100% sure how much it helps

Shall spend a bit more time on the prototype and see if it'll work with a reasonable amount of effort

djmitche:
Regarding async -- yeah, it still has some rough edges!  I think at least some Future implementations are `#[must_use]`, which would at least warn in this case, but perhaps not all of them.

Another option to sending sql and getting back values might be to send a closure and get back that closure's result.  But if `types::Value` is working, then that's great :)

dbr:
Finally got back to this!

I now have something (inefficient) working by creating a new connection per query - so all the tests are passing and appears to function correctly.. however creating a new rusqlite connection per query on the server isn't exactly ideal.

Next I need to:

1. Update branch to be inline with `main`
2. Integrate the async-wrapper prototype

dbr:
> Regarding async -- yeah, it still has some rough edges! I think at least some Future implementations are #[must_use], which would at least warn in this case, but perhaps not all of them.

From vague memory, the problem was I *did* use the future, but it was used by something which almost immediately got dropped, like `rt.spawn_blocking(thefuture)` instead of `let _ = rt.spawn_blocking(thefuture)`

> Integrate the async-wrapper prototype

This has proved quite challenging - I've tried a few things:

1. Just using r2d2 + r2d2_sqlite connection pool. The pool is send+sync.. but shoving this into the existing SqliteStorage + Txn structs proved akward:

I got a `rusqlite::Transaction` from the `r2d2::PooledConnection<...>`, but I can't store both of these in the `Txn` instance because of lifetime related error

I think this is solvable the same way the `sync-server/src/storage/inmemory.rs` deals with the MutexGuard (but I only noticed that after trying next things)

2. Async wrapper around rusqlite - a `Connection` struct which reads from a mpsc queue in a 'main loop", it has a collection of `Transaction` structs, each of these have an equivalent "handle" struct which uses message passing to get data to/from the main loop

Initially I tried having this be "all async" using Tokio, then realized that a) spawn_blocking requires send+sync also, and b) the storage API isn't async - I could make the main rusqlite-processing loop use `std::sync::mpsc` things.. but this seemed like a lot of effort not not much gain (the API loses almost any benefit from being async if the storage blocks)

3. So I then tried making the storage API async, using [`async_trait`](https://github.com/dtolnay/async-trait) to have `pub(crate) trait StorageTxn { async fn get_client(&mut self, client_key: Uuid) -> anyhow::Result<Option<Client>>; ...}` etc

This seemed to get me closer, however the main difficulty is getting the data back in a nice way. Sending a closure and using this to get data back in right format is the neatest (using `rusqlite::types::Value` makes the "client" side of things very clunky)

Problem is putting the closure in a `enum` to send via message passing is.. not obvious - the `query_row` looks something roughly like:

```rust
    pub(crate) async fn query_row<F>(
        &self,
        sql: String,
        params: & [&dyn rusqlite::types::ToSql],
        f: F,
    ) -> Result<T, rusqlite::Error> 
    where F: FnOnce(rusqlite::Row) -> T {
        // Make return channel
        let (result_send, result_recv) = oneshot::channel();
        // Send request
        self.sender
            .send(TransactionMessage::QueryRow(sql, params, f, result_send)).await?;
        // Wait on returned value
        Ok(result_recv.await?)
    }
```

..but how to define the `QueryRows` enum variant is beyond me (the params can be turned into `Vec<rusqlite::types::Value>` easily, but it's not clear at all how the closure with generic type parameter would work)

..at which point I realized this was probably making things too complicated for what it needs to be, and I'm tempted to go back to the r2d2 connection pool and fix the lifetime thing with the extra level of Struct as per the `inmemory.rs`

Any thoughts on best way to approach this?

djmitche:
So first, I think landing something inefficient but functional is not a bad first step!  Cleaning up this PR as-is by resolving conflicts is probably enough to land it and release 0.4.0.

Making the storage API async is a good idea anyway, since the callers of these methods in sync-server are already async.  That's not the case for the Storage trait in the `taskchampion` crate -- embedding async Rust into other language is likely to be a nightmare, so we should leave that trait sync.

It does seem like the "simpler" way to solve the connection pooling issue, and the one that will be a better guide for whoever implements a postgres backend, would be with `r2d2`.  I'm not sure what the lifetime issue looked like, but it seems more difficult than for InMemory: you would need to store both a PooledConnection and a Transaction, with the Transaction referencing the PooledConnection.  That's not easy to do safely.  I did some experimentation [here](https://gist.github.com/djmitche/33353e70a967d6eb5c82f914abc6026f) and asked about it in `#rust:matrix.org`.

Again separating the sync-server storage trait from that in `taskchampion`, maybe the right fix is to model storage with three levels as well -- Storage / StorageConnection / StorageTxn.  Or, we could drop the idea of a "transaction object" since we don't need it in the sync server, and instead just have a single Storage trait with methods that execute the whole transaction before returning.

In any case, let's get this landed.  We can open new issues to revisit (a) making the sync-server storage API async and (b) implementing connection pooling with the sqlite implementation.  Neither of those will change the on-disk format or any public APIs, so they don't need to be aligned with a release.

djmitche:
..and separating those two projects will probably be easier :)

dbr:
Yep sounds like a good plan!

dbr:
Made separate tickets for the async-storage (https://github.com/taskchampion/taskchampion/issues/277) and SQLite connection pooling (https://github.com/taskchampion/taskchampion/issues/278)

sync-server/src/storage/sqlite.rs
```diff
@@ -0,0 +1,295 @@
...
+    fn txn<'a>(&'a self) -> anyhow::Result<Box<dyn StorageTxn + 'a>> {
+        let con = self.new_connection()?;
+        let t = Txn { con, txn: None };
+        Ok(Box::new(t))
+    }
+}
+
+struct Txn<'t> {
+    con: Connection,
+    txn: Option<rusqlite::Transaction<'t>>,
```
djmitche, line 97: It looks like this is never set.  That would cause `commit` to panic, but `commit` is never called.  I think all of that is fine, since the server storage API doesn't need to bundle multiple method calls into a transaction.  So, it just needs a bit of cleanup (removing this field and the `commit` method).

dbr: Oh quite right, was a left-over from trying to get the connection/transaction stuff working I think!

taskchampion/src/server/local.rs
```diff
@@ -1,104 +1,141 @@
...
-use kv::{Bucket, Config, Error, Integer, Serde, Store, ValueBuf};
+use anyhow::Context;
+use rusqlite::params;
+use rusqlite::types::{FromSql, ToSql};
+use rusqlite::OptionalExtension;
 use serde::{Deserialize, Serialize};
 use std::path::Path;
 use uuid::Uuid;
 
+// FIXME: Duplicated
```
djmitche, line 12: Duplicated with taskchampion/src/storage/sqlite.rs?  That should be relatively easy to fix..

taskchampion/src/storage/sqlite.rs
```diff
@@ -1,355 +1,346 @@
...
+    fn get_next_working_set_number(&self) -> anyhow::Result<usize> {
+        let t = self.get_txn()?;
+        let result: Option<usize> = t
+            .query_row("SELECT COALESCE(MAX(id), 0) FROM working_set", [], |r| {
+                r.get(0)
+            })
+            .optional()
+            .context("Getting highest working set ID")?;
+
+        Ok(result.unwrap_or(0) + 1)
```
djmitche, line 122: just for clarity:
```suggestion
            .query_row("SELECT COALESCE(MAX(id), 0) + 1 FROM working_set", [], |r| {
                r.get(0)
            })
            .optional()
            .context("Getting highest working set ID")?;

        Ok(result.unwrap_or(0))
```

taskchampion/src/storage/sqlite.rs
```diff
@@ -1,355 +1,346 @@
...
+            [&StoredUuid(uuid)],
+            |x| x.get(0),
+        )?;
+        if count > 0 {
+            return Ok(false);
         }
+
+        let data = TaskMap::default();
+        t.execute(
+            "INSERT INTO tasks (uuid, data) VALUES (?, ?)",
```
djmitche, line 161: Could maybe just use `INSERT OR IGNORE` here?

taskchampion/src/storage/sqlite.rs
```diff
@@ -1,355 +1,346 @@
...
 
     fn set_operations(&mut self, ops: Vec<Operation>) -> anyhow::Result<()> {
-        let numbers_bucket = self.numbers_bucket();
-        let operations_bucket = self.operations_bucket();
-        let kvtxn = self.kvtxn();
-
-        kvtxn.clear_db(operations_bucket)?;
+        let t = self.get_txn()?;
+        t.execute("DELETE FROM operations", [])
+            .context("Clear all existing operations")?;
```
djmitche, line 269: Maybe also reset the indexing? https://www.designcise.com/web/tutorial/how-to-reset-autoincrement-number-sequence-in-sqlite#deleting-the-table-number-sequence

taskchampion/src/storage/sqlite.rs
```diff
@@ -1,355 +1,346 @@
...
+                Ok((id, uuid.0))
+            })
+            .context("Get working set query")?;
 
-        let mut res = Vec::with_capacity(next_index as usize);
-        for _ in 0..next_index {
-            res.push(None)
+        let rows: Vec<Result<(usize, Uuid), _>> = rows.collect();
+        let mut res = Vec::with_capacity(rows.len());
+        for _ in 0..self.get_next_working_set_number().context("HUh")? {
```
djmitche, line 291:
```suggestion
        for _ in 0..self.get_next_working_set_number().context("Getting working set number")? {
```

djmitche (approved):
Apologies for taking so long to review this!

This looks good -- feel free to take or reject the suggestions below.

taskchampion/src/storage/sqlite.rs
dbr: That would make it hard to have the return-true-if-already-exist return value I think (but I may be misunderstanding?)

taskchampion/src/server/local.rs
dbr: Ah true, I think at the time I (half-)thought "server.rs -> sync-server -> different crate" so it would have had to be `pub`.. but both are in the same crate :roll_eyes: Fixed!

taskchampion/src/storage/sqlite.rs
dbr: Yep makes sense - I think that'll might work better also (since if there is no results, the old code would have returned `1` for first ID). Also renamed the terrible `result` variable to something more descriptive

taskchampion/src/storage/sqlite.rs
dbr: Makes sense - done!

taskchampion/src/storage/sqlite.rs
djmitche: Ah, that makes sense -- I think postgres has a way to return that information, but figuring that out for sqlite would not be worthwhile.

djmitche (approved):
Sheep It!

dbr:
Moo-erged in main branch, added changelog entry, shall merge when CI is appeased

djmitche:
once this merges I think we're done with 0.4.0?

dbr:
I was just creating a new issue about making 0.4.0 release! Yep I think this is the last issue left in that milestone :partying_face:

PR #218, opened 2021-04-28
Add CGI assets

savchenko:
Logo plus square and rounded icons.

- Minor fine-tuning.
- Ribbon is present throughout.
- Exported in most common resolutions.

Closes https://github.com/taskchampion/taskchampion/issues/207


djmitche (commented):
Looks good!

What is "CGI" in this context?

Also, would it make sense to include the vector source? And if scripted, a means for creating the PNGs from the vector?

savchenko:
> Looks good!
> 
> What is "CGI" in this context?
> 
> Also, would it make sense to include the vector source? And if scripted, a means for creating the PNGs from the vector?

ComputerGeneratedGraphics (CGI).

Will share the source vectors in chat. Creating SVG of a civilized size would be a non-trivial exercise due to some of the effects used.

djmitche:
Are those source vectors something we can't distribute?  If they're super-sized, maybe another repo would be appropriate.  I've been caught out in other projects (buildbot) by not having originals for logos and having to reconstruct them by hand.

savchenko:
> Are those source vectors something we can't distribute? If they're super-sized, maybe another repo would be appropriate. I've been caught out in other projects (buildbot) by not having originals for logos and having to reconstruct them by hand.

We probably can, but they are created in proprietary software and are ~3mb blobs in total. I have sent both in chat.

dbr:
Although the file size isn't huge, given:
1. we don't necessarily need all these sizes/variants (some might be only useful in, say, a phone application which would most certainly be a separate project/repo)
2. we might need some additional specific sizes or formats
3. there might be tweaks needed to the icons (which would involve recreating them all)
4. git doesn't allow things to be removed from history

..then I'd be tempted to put these all in a separate taskchampion-graphics repo

If we do that, I think it's good to include the source files in the same repo (even if it is in a proprietary format)

savchenko:
All assets including the source files are uploaded to https://github.com/taskchampion/visuals

I can't update TC organisation icon though. @djmitche , I would recommend using `icon_square_512.png` for it.

dbr:
I'd possibly suggest prefixing the repo with `taskchampion-` to avoid `git clone` or forks outside of TC'c org giving a folder just called `visuals`, but I don't have a strong opinion on that. Aside from that, looks good to me!

djmitche:
> I can't update TC organisation icon though. @djmitche , I would recommend using `icon_square_512.png` for it.

Done!

PR #221, opened 2021-05-01
Require a filter be specified for modifications

### LINKED ISSUE #177: Do not modify all tasks by default; require `*` for that
As per https://github.com/djmitche/taskchampion/discussions/153

`task 1 start` marks one task as active. Good.

Expected behaviour: `task start 1` should be equivalent

Current (v0.3) behavior: it finds all tasks matching empty filter, then modifies them with description 1 - essentially clobbering the description of all tasks with `"1"`

djmitche:
This filter can either be `*` or some "real" filter.  But an empty set
of arguments no longer automatically matches all tasks.

Fixes #177.

### DIFF
--- a/cli/src/argparse/filter.rs
+++ b/cli/src/argparse/filter.rs
@@ -1,8 +1,13 @@
-use super::args::{arg_matching, id_list, minus_tag, plus_tag, status_colon, TaskId};
+use super::args::{arg_matching, id_list, literal, minus_tag, plus_tag, status_colon, TaskId};
 use super::ArgList;
 use crate::usage;
 use anyhow::bail;
-use nom::{branch::alt, combinator::*, multi::fold_many0, IResult};
+use nom::{
+    branch::alt,
+    combinator::*,
+    multi::{fold_many0, fold_many1},
+    IResult,
+};
 use taskchampion::Status;
 
 /// A filter represents a selection of a particular set of tasks.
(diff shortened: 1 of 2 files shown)


dbr:
Change looks good to me, although I just added comment to #177 which relates to this (`*` might be annoying as it's shell glob syntax, maybe `all` instead?)

It would be nice if `ta done` gave a helpful "did you mean: `task all done` or `task 1 done`" type error message, but I imagine that is probably a thing for the future

djmitche:
I suspect this is a very niche use-case, so I don't think it needs such a common word as "all".  But you're right about the shell expansion.

Maybe I'll switch this around so that `ta all done` will work, but `ta all list` will not -- that is, where I use `Filter::parse1`, I'll allow `all`, but not where I use `Filter::parse0`.  That at least limits the chance for ambiguity.

dbr:
Played around with this and it's much better behaviour! I've been avoiding actually using TC because with the old `ta done 1` clobbering everything was too easy

Main issue I can see is the error when doing the wrong thing is a bit unclear:
```
$ cargo run --bin ta -- done 1
report `done` not defined
```

Ideally that'd say "missing required filter before 'done', e.g '1 done'" or similar. Maybe this is an improvement for a separate, future, PR?

djmitche:
Yeah, please file a new issue for that.  I think CLI error handling in general needs some thought.

Issue #164, opened 2021-03-26
CLI command to set config options


dbr:
As discussed in https://github.com/djmitche/taskchampion/discussions/152

It would help with writing good user-guides to have a CLI which can set config options, maybe something like

    task config set sync-server-origin http://example.com

or maybe more locally specific

    task sync set-url http://example.com

or something along those lines.

Justification being that "run this command to set the server" is a lot friendly than "edit the config at [platform dependent location], then in this section, add the line [...etc]"

Git is a good reference for this - e.g the commands for setting name/email are presented in error messages from git itself, remote URL's are presented by hosting services like Github.

djmitche:
This seems quite common.  I think we'd need to pick another config language, though -- I don't think the YAML parser preserves enough state (comments, whitespace) to "edit" a file without losing the user's formatting.  I think TOML can do this?  At least, `cargo add` edited my `Cargo.toml` without losing comments..  And TOML seems to be getting more popular as a config format.

dbr:
I'm in favour of using something which isn't YAML (popular but *giant* spec leading to things like [this website to help people with how to format a multi-line string](https://yaml-multiline.info/), among many other issues)

TOML would be my vote, although for how widely use it is I found the main `toml` crate surprisingly under-developed - I don't think it [currently supports preserving comments](https://github.com/alexcrichton/toml-rs/issues/83), and I've run in to other things like it being [sensitive to the order fields are defined](https://github.com/alexcrichton/toml-rs/issues/336). `cargo-edit` seems to use [this toml_edit](https://github.com/ordian/toml_edit) crate instead

I have also looked into [ron](https://github.com/ron-rs/ron) which seems interesting, as is [hcl](https://github.com/hashicorp/hcl) - although the former is fairly Rust-only, and the latter hasn't really caught on (and now I think about it, I'm not sure I found a Rust implementation of it.. so.. probably not too useful a suggestion!). There is also https://github.com/google/json5format which is JSON with support for preserving comments and some other changes, but seems quite early stages

Possibly `toml_edit` with the hope we can switch to the main `toml` crate someday when https://github.com/alexcrichton/toml-rs/issues/83 gets more attention?

savchenko:
TOML sounds like a good balance. Given how central configuration is, do you think this can be added as early as in [v0.4.0](https://github.com/taskchampion/taskchampion/milestone/1)?

djmitche:
Yes, that makes sense.

PR #224, opened 2021-05-02
Switch to TOML for configuration

djmitche:
Refs #164.

This doesn't (yet) support dynamic editing of the config file, but switches the format to TOML.


cli/src/invocation/mod.rs
```diff
@@ -100,35 +100,33 @@ pub(crate) fn invoke(command: Command, settings: Config) -> anyhow::Result<()> {
...
+fn get_server(settings: &Settings) -> anyhow::Result<Box<dyn Server>> {
     // if server_client_key and server_origin are both set, use
     // the remote server
-    let config = if let (Ok(client_key), Ok(origin)) = (
-        settings.get_str("server_client_key"),
-        settings.get_str("server_origin"),
+    let config = if let (Some(client_key), Some(origin), Some(encryption_secret)) = (
+        settings.server_client_key.as_ref(),
+        settings.server_origin.as_ref(),
+        settings.encryption_secret.as_ref(),
```
dbr, line 117: Is it an issue that previously the encryption key missing was an error if the client key+origin were set, but now it will quietly ignore the remote server?

I guess I would expect that if the `server_origin` is set, the client key and encryption secret are both required (and it's an error to omit them)

Probably something to split into a new issue and worry about later, but I wonder if it'd make sense to have the entire server address as a single string (along the lines of the HTTP `key:secret@serveradd` or Postgres connection strings etc). Would be easier for users to copy one string, and nicer for the code to handle

dbr:
LGTM!

cli/src/invocation/mod.rs
djmitche: I'm not too worried about the change (I suspect it was an oversight when I added encryption).  In general this "if-it's-present" is not a great UI.  But, let's solve that when the server bits are more stable.

PR #225, opened 2021-05-03
Support multiple exit codes

### LINKED ISSUE #217: CLI exit codes
As per https://github.com/taskchampion/taskchampion/pull/212#discussion_r621458155, `ta` should use meaningful return codes. Current list under proposal:

- `0` No errors, normal exit.
- `1` Generic error.
- `2` Never used to avoid conflicts with Bash.
- `3` Unable to execute with the given parameters.
- `4` I/O error.
- `5` Database error.

djmitche:
..with more specific error enums.

Still a draft because for the most part this will never return 4 or 5 (IO or Database errors).

Fixes #217.


dbr:
I don't feel strongly about it, but I'm not too sure about just having `Result<..., Error>` in methods as it's not too clear what Error type it is (e.g `std::io::Error`?). Would it maybe make sense to re-export Error from the crate root and use `Result<..., crate::Error>`? Or maybe name it `CliError`?

Aside from that, looks good!

djmitche:
redrafted to use `crate::Error` as suggested!

dbr (approved):
🥳

PR #228, opened 2021-05-05
Support editing config file from the command line

### LINKED ISSUE #164: CLI command to set config options

djmitche:
Fixes #164.


dbr:
This looks like a great change :partying_face: The docs for configuring the sync server seem a lot easier to follow along given this change

cli/src/invocation/cmd/config.rs
```diff
@@ -0,0 +1,62 @@
...
+    match config_operation {
+        ConfigOperation::Set(key, value) => {
+            let filename = settings.set(&key, &value)?;
+            write!(w, "Set configuration value ")?;
+            w.set_color(ColorSpec::new().set_bold(true))?;
+            write!(w, "{}", &key)?;
+            w.set_color(ColorSpec::new().set_bold(false))?;
+            write!(w, " in ")?;
+            w.set_color(ColorSpec::new().set_bold(true))?;
+            write!(w, "{:?}.\n", filename)?;
```
dbr, line 19: Hmm, not sure I agree with the clippy lint here (maybe a separate `write!(w, "\n")` would be clearer and appease the lint?)

However I wonder if it's worth investigating a neater way to do the coloured text formatting as `write!(w, "Set configuration value {} in {:?}", ...)` is drastically easier to read than the version with the coloured output

dbr: Split this into #229 as it's unrelated to this change really

djmitche: I'm OK with appeasing clippy, though :)

dbr (approved):
LGTM!

PR #230, opened 2021-05-07
Document environment variables all in one place

### LINKED ISSUE #100: document environment variables
Add environment variables to the list of things that `crate::usage` knows how to document, and include the env vars used to find the config files as well as NO_COLOR and TERM for https://crates.io/crates/termcolor.

djmitche:
With the switch to TOML, we no longer support overriding specific configuration with environment variables, so there's no good reason to include this in the `usage` module's functionality.

Fixes #100.


dbr:
Just checking I understand properly:

- The old settings setup allowed user you override `such_and_such` a setting using `$SUCH_AND_SUCH`
- New TOML implementation doesn't allow that
- Therefore #100 is no longer directly relevant, we just need to document the few env-vars used by termcolor etc

Assuming that is right, LGTM.

PR #231, opened 2021-05-07
Fix a working set rebuild bug

djmitche:
I ran into an issue in day-to-day use where the code tried to set working-set index 0 to None.  First, there is no working set index 0; and second, setting a working-set index to None that is already None should be valid.  So, all of that is fixed, with some extra checks.


dbr:
Hmm, I wonder if it'd help avoid these types of bugs to have a `struct WorkingSet` which implements `std::ops::Index` etc?

Maybe a future thing - this change looks good for now. I should also copy these test cases over to the SQLite backend branch

djmitche:
We have a `struct WorkingSet`, but it doesn't implement Index -- that's a good idea!  I think this particular function wouldn't be helped, though, as it's handling the "raw" data.

PR #236, opened 2021-05-12
Generate usage documentation

djmitche:
This introduces a `generate` binary that updates the documentation based on the built-in usage.  This will ensure that the docs match `ta help`.


dbr:
Asking mostly because I'm planning to do something similar at some point [for auto-generating screenshots in mdbook](https://github.com/imgui-rs/imgui-rs/pull/482), did you investigate at all the mdbook "preprocessor" feature to implement this?

https://rust-lang.github.io/mdBook/for_developers/preprocessors.html

Would be neater in that the source files remain "clean", and it also uses the mdbook parser. However this may be excessive in this case, and also maybe not quite as nice since `reports.md` etc wouldn't contain anything but the placeholder, so only the generated mdbook would be readable :thinking:

djmitche:
I didn't know about that functionality -- I'll take a look!

djmitche:
Yeah, this looks like a good choice, but will be a significant refactor.  Thanks!

djmitche:
Hopefully that's green this time.  Thanks for the tip!  This is much nicer.

README.md
```diff
@@ -33,3 +33,8 @@ There are three crates here:
  * [taskchampion-cli](./cli) - the command-line binary
  * [taskchampion-sync-server](./sync-server) - the server against which `task sync` operates
 
+## Code Generation
+
+The `generate` binary generates files that are checked into the repository.
```
dbr, line 38: This is section is outdated by `usage-docs` I think?

build-docs.sh
```diff
@@ -0,0 +1,26 @@
...
+if ! [ -f "docs/src/SUMMARY.md" ]; then
+    echo "Run this from the root of the repo"
+    exit 1
+fi
+
+# build the latest version of the mdbook plugin
+cargo build -p taskchampion-cli --features usage-docs --bin usage-docs
+
+if ! [ -d ./docs/tmp ]; then
+    git worktree add docs/tmp gh-pages
```
dbr, line 16: Might be worth adding a few comments to what each of these are doing - all makes perfect sense, just a bit hard to determine what each line is doing on first glance (e.g "make working copy of gh-pages in doc/tmp/", "remove old generated docs" etc)

dbr (changes requested):

dbr:
One reasonable looking clippy lint left! :grimacing: Also few very minor notes, but otherwise looks really good - the mdbook processor thing seems quite neat!

dbr (approved):

Issue #241, opened 2021-05-19
More descriptive --version flag


savchenko:
It would be useful to show the commit ID if TaskChampion is built from HEAD.
Currently `--version` always returns the version number. Thoughts?

djmitche:
I'm not sure how to do that.. @dbr?  Is there a way to do so in `build.rs`?

savchenko:
I don't speak Rust, but perhaps `git describe --tags --always` would do?

dbr:
Somewhat cumbersome but you can set env-vars in a `build.rs` script,
https://stackoverflow.com/questions/43753491/include-git-commit-hash-as-string-into-rust-program
..which is done here:
https://github.com/taskchampion/taskchampion/pull/242

`git describe --tags` is interesting (didn't know about that), but I think probably still best to use the version Cargo package version combined with `git rev-parse --short HEAD`, otherwise the git tags could get out of sync with what is pushed to http://crates.io which would be pretty confusing

PR #242, opened 2021-05-21
Include git HEAD rev in version output

dbr:
As per #241

Output is something like:

```
TaskChampion 0.3.0 (git rev: 1d6c3aa)
```


djmitche (approved):
Nice!  This will definitely help with QA :)

PR #245, opened 2021-05-22
add 'ta config path'

### LINKED ISSUE #138: Make a `task config` command to dump the configuration
..including defaults.

djmitche:
Fixes #138

### DIFF
--- a/cli/src/invocation/cmd/config.rs
+++ b/cli/src/invocation/cmd/config.rs
@@ -19,6 +19,13 @@ pub(crate) fn execute<W: WriteColor>(
             writeln!(w, "{:?}.", filename)?;
             w.set_color(ColorSpec::new().set_bold(false))?;
         }
+        ConfigOperation::Path => {
+            if let Some(ref filename) = settings.filename {
+                writeln!(w, "{}", filename.to_string_lossy())?;
+            } else {
+                return Err(anyhow::anyhow!("No configuration filename found").into());
+            }
+        }
     }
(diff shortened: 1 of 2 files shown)


cli/src/invocation/cmd/config.rs
```diff
@@ -19,6 +19,13 @@ pub(crate) fn execute<W: WriteColor>(
             writeln!(w, "{:?}.", filename)?;
             w.set_color(ColorSpec::new().set_bold(false))?;
         }
+        ConfigOperation::Path => {
+            if let Some(ref filename) = settings.filename {
+                writeln!(w, "{}", filename.to_string_lossy())?;
+            } else {
+                return Err(anyhow::anyhow!("No confiuration filename found").into());
```
dbr, line 26: Typo in `configuration`

dbr (changes requested):
Looks good but for the typo!

dbr:
BTW I've not played around with it yet but I found https://docs.rs/camino/1.0.4/camino/ which seems interesting alternative to `to_string_lossy` and lots of `Result<..>` handling with file paths

> camino is an extension of the std::path module [...] like the standard library’s PathBuf and Path types, except they are guaranteed to only contain UTF-8 encoded data

djmitche:
I'm not terribly worried about the `_lossy` bit.  I suspect the number of people who will be using non-utf8 filenames with taskchampion (which doesn't do much with files) is approximately zero :)

PR #254, opened 2021-05-30
Define and test an MSRV

### LINKED ISSUE #216: Minimum supported Rust version
#212 mentions targeting only stable Rust, however in terms of making it easier to package on Linux distros, it might make sense to ensure we can build TC with a specific Rust version (tested via CI)

https://repology.org/project/rust/badges

Based on that I'd say 1.47 because:

- Debian Testing is on 1.48
- CentOS is on 1.47
- Ubuntu 16.04 through 20.10 is on 1.47
- Alpine Linux 3.13 is also on 1.47 (may be useful for docker container for sync-server)
- The rest I'm either not familiar with or are on 1.51

I don't think there's too much we'd miss in Rust 1.48-1.51:

- https://blog.rust-lang.org/2020/11/19/Rust-1.48.html
- https://blog.rust-lang.org/2020/12/31/Rust-1.49.0.html
- https://blog.rust-lang.org/2021/02/11/Rust-1.50.0.html
- https://blog.rust-lang.org/2021/03/25/Rust-1.51.0.html

Major changes in those versions are const-generics (not too useful for TC) and nicer linking in rustdocs (we can still this as we would only be maintaining support for a specific version of Rust, not necessarily actually using it for builds). Main trouble will likely be dependencies relying on newly stabilized API's

djmitche:
Fixes #216.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -1,0 +1,43 @@
+name: taskchampion
+
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+    types: [opened, reopened, synchronize]
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+
+    strategy:
+      matrix:
+        rust:
+          - "1.47" # MSRV
+          - "stable"
+
+    name: "Test - Rust ${{ matrix.rust }}"
+
+    steps:
+      - uses: actions/checkout@v1
+
+      - name: Cache cargo registry
+        uses: actions/cache@v1
+        with:
+          path: ~/.cargo/registry
+          key: ${{ runner.os }}-${{ matrix.rust }}-cargo-registry-${{ hashFiles('**/Cargo.lock') }}
+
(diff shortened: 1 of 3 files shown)


.github/workflows/tests.yml
```diff
@@ -0,0 +1,43 @@
+name: taskchampion
```
dbr, line 1: Currently the "Checks" tab has two sections titled "taskchampion", would possibly be neater if this said "tests" or similar?

There's also two checks titled "Clippy" which seems slightly odd (not sure if they are doing different things?)

djmitche: I kind of don't understand the Actions UI and where it gets names from.  I'm happy to see this addressed in another PR.

djmitche:
Oh, I thought that was approved-with-comments, sorry!

dbr:
> Oh, I thought that was approved-with-comments, sorry!

It was I think (or at least that's what I meant to click!)

PR #260, opened 2021-06-01
produce Tag instances in the parser

### LINKED ISSUE #111: Parse CLI args as Tag earlier
In `crate::argparse::args`, `plus_tag` and `minus_tag` should both return a `Tag`.  Then, Filter and Modify should handle those as already-parsed values.  Currently tags are maintained as strings until `invocation`, at which point it's kind of too late to handle any errors converting from String to Tag.

This is a good internal-refactor issue.

djmitche:
Fixes #111

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1630,9 +1630,9 @@ checksum = "7ffc5c5338469d4d3ea17d269fa8ea3512ad247247c30bd2df69e68309ed0a08"
 
 [[package]]
 name = "mdbook"
-version = "0.4.8"
+version = "0.4.9"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "fed4060ccf332a0479df37e84c8435ad20be737d5337c3a90fa1b3b0d480a3a0"
+checksum = "aeb86d199d0c1e8d41f3a9e9b0ba8639d6951a10043129159809ac9c18e3ce05"
 dependencies = [
  "ammonia",
  "anyhow",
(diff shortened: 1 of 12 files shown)


dbr (approved):
The clippy lint seems reasonable thing to fix, but otherwise looks good to merge I think!

djmitche:
(rebased and updated)

djmitche:
Hm, I don't know why the docs are failing..

dbr:
That is very odd.

I printed out `stdin` inside the usage-docs command and it looks pretty reasonable - if I cull all the docs content so it's short, it just says something like:

```json
[{"root":"/Users/dbr/code/taskchampion/docs","config":{"book":{"authors":["Dustin J. Mitchell"],"language":"en","multilingual":false,"src":"src","title":"TaskChampion"},"output":{"html":{"default-theme":"ayu"}},"preprocessor":{"usage-docs":{"command":"target/debug/usage-docs"}}},"renderer":"html","mdbook_version":"0.4.9"},{"sections":[]}]
```

..which is all in order

Ah, I think it's a bug in mdbook 0.4.8 (version in TC's Cargo.lock) - if I update to 0.4.9 it seems happy again

This change likely fixed it: https://github.com/rust-lang/mdBook/compare/7e01cf9...d325c60#diff-030bbc3b81baed9210053c583963f03bdcbcd802c2f065254400ffe6266306d6L233

djmitche:
I'll squash this if it's green.

PR #262, opened 2021-06-05
Remove path from ta command in usage text

dbr:
Minor thing but makes the help text a lot less visually dense (when doing `cargo run` at least)

```
TaskChampion subcommands:
  ta version
    Show the version of the TaskChampion binary
  ta config set <key> <value>
    Update Taskchampion configuration file to set key = value
```

..which was previously:

```
TaskChampion subcommands:
  target/debug/ta version
    Show the version of the TaskChampion binary
  target/debug/ta config set <key> <value>
    Update Taskchampion configuration file to set key = value
```


djmitche (approved):
Looks good!  This will catch the case where someone changes the binary name, while still providing readable output.

PR #263, opened 2021-06-05
Tidy up actions names

dbr:


dbr:
So I think this is about as good as I can get it. Only main weirdness is the "Clippy Results" thing which produces the inline-annotations (from https://github.com/actions-rs/clippy-check ) seems to appear under a random section and there's no obvious way to control this

dbr:
Ah the weirdness is a known bug, https://github.com/actions-rs/clippy-check/issues/45

djmitche (approved):
Looks good.  Maybe we can also remove "Test - " from the test actions?  This is also fine by me as-it-is.

djmitche:
Did you want to merge this?

dbr:
Woops, yep! Thanks for the reminder

Issue #253, opened 2021-05-30
Use a custom `nom` error type


djmitche:
Errors from the command-line parser are _terrible_ right now.  Let's improve those with a custom error type and thoughtful use to generate useful error messages when something goes wrong.

dbr:
This likely supersedes #246

djmitche:
Yes, but let's keep that around as a specific thing to fix in the process.

PR #265, opened 2021-06-05
Treat partially-matched arguments as an error

djmitche:
This is correct, but leaves a lot of room for improvement in terms of error handling (#253).


dbr (approved):
LGTM!

PR #266, opened 2021-06-06
Add support for synthetic tags

### LINKED ISSUE #80: Implement synthetic tags
TaskWarrior has a bunch of synthetic tags like +WAITING and +PENDING.  We could do the same thing pretty easily.  They should be synthesized entirely within the replica, rather than stored in the task DB.

djmitche:
Notably, nothing changes for the CLI -- this is entirely within the taskchampion crate, and everything outside just sees all-caps tags.

Fixes #80.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -826,8 +826,8 @@ dependencies = [
  "serde",
  "serde_derive",
  "serde_json",
- "strum",
- "strum_macros",
+ "strum 0.18.0",
+ "strum_macros 0.18.0",
 ]
 
 [[package]]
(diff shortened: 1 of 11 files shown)


djmitche:
BTW, this is a re-implementation of https://taskwarrior.org/docs/tags.html, so that might provide context for decisions made here.

dbr:
Change looks good, although I'm not too sure about the `dialoger` library's prompt - in a narrow terminal on macOS Terminal.app it doesn't wrap text, and repeats the line on input:

Screen Shot 2021-06-09 at 12 00 04 pm

dbr:
Seems good to me!

My initial thought was maybe the `SyntheticTag` etc should be in the public API, to allow GUI's and such to do specific things with them (e.g display them differently, select them from a dropdown list when filtering etc) - but I'm not too sure about that, probably best to just wait and expose it later if required

djmitche:
> probably best to just wait and expose it later if required

I think so, too.  At the moment, such features could be implemented by just string-matching.

PR #267, opened 2021-06-07
Add confirmation prompts for modifications of lots of tasks

### LINKED ISSUE #179: Confirmation prompts
There are some batch operations where a confirmation may be helpful, e.g `task start` may want to prompt before marking every task as active

If undo (#92 ) is implemented, some of these could potentially be removed, and instead the output just inform the user that undo exists, something along the lines of "Note: Marking 1120 tasks as active - if this was not intended, run 'task undo' immediately to reverse this operation"

djmitche:
Fixes #179.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -666,6 +666,21 @@ dependencies = [
 ]
 
 [[package]]
+name = "console"
+version = "0.14.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "3993e6445baa160675931ec041a5e03ca84b9c6e32a056150d3aa2bdda0a1f45"
+dependencies = [
+ "encode_unicode",
+ "lazy_static",
+ "libc",
+ "regex",
+ "terminal_size",
+ "unicode-width",
+ "winapi 0.3.9",
+]
+
+[[package]]
 name = "const_fn"
 version = "0.4.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -758,6 +773,18 @@ dependencies = [
  "proc-macro2",
  "quote",
  "syn",
+]
+
+[[package]]
+name = "dialoguer"
+version = "0.8.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "c9dd058f8b65922819fabb4a41e7d1964e56344042c26efbccd465202c23fa0c"
+dependencies = [
+ "console",
+ "lazy_static",
+ "tempfile",
+ "zeroize",
 ]
 
 [[package]]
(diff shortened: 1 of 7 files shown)


dbr:
(posted on the right PR this time!)

Change looks good, although I'm not too sure about the `dialoger` library's prompt - in a narrow terminal on macOS Terminal.app it doesn't wrap text, and repeats the line on input:

Screen Shot 2021-06-09 at 12 00 04 pm

It works okay in iTerm2, haven't checked on any Linux terminals yet

djmitche:
It would be the same on any terminal.  I'll switch back to the simple theme :)

djmitche:
(and fix the test)

dbr:
That works better, but still a little weirdly - on input, the two line prompt gets overwritten and repeated:

Screen Shot 2021-06-10 at 9 54 31 am

I wonder if we should:

1. Merge this PR pretty much as is
2. Then, fix this problem in upstream library - doesn't seem like there is many other options, and otherwise the prompt seems quite nice
3. Possibly make the prompt shorter by moving the note about the setting to after the prompt, maybe something like:

        ✔ Operation will modify at least 6 tasks; continue? · no
        Cancelled (note: prompted because config "modification_prompt_count" is 3)

    Would be nice if `ta config set ...` could be worked into that prompt, but gets quite wordy.. but maybe that doesn't matter if we:

4. Store the `modification_prompt_count` as `Option<u64>`. If `None` then the prompt appears and is treated as the default value of 3. If `Some(...)` then the prompt is suppressed for neatness (since the user is hopefully aware of the config option as it's been changed)

    This be useful for other options too - it can be helpful to tell if a setting is at the default, or if the user has explicitly configured it (which might or might not be to the same default value).

    Basically something like `loaded_settings.thingo.unwrap_or(default_settings.thingo)` instead of `loaded_settings = Settings.default(); loaded_settings.update(...)`. This also means it's easy to then store only the modified settings in the config file, which in turn makes it easier if we ever want to update default settings

dbr:
Also, pedantically, I wonder if the setting should be named `modification_count_prompt` (since it's a prompt based on the modification count, not a count of modification prompts)

dbr:
Also weird failure in the docs build,

> Warning: This mdbook preprocessor was built against version 0.4.9 of mdbook, but we're being called from version 0.4.10

djmitche:
> Also weird failure in the docs build,
> 
> > Warning: This mdbook preprocessor was built against version 0.4.9 of mdbook, but we're being called from version 0.4.10

That's fixed on main.

djmitche:
I suspect option 2 is difficult -- detecting wrapping in terminal UIs is rather hard.  I did the other things though!

djmitche:
Thanks for the reviews here -- it's definitely improved the PR so far :)

dbr:
Looks good! Wrapping weirdness still happens, but now only in an unreasonably narrow terminal, so not worth worrying about

I also noticed `ta config set` doesn't work if the config doesn't already exist, but that is definitely a separate issue!

dbr (approved):

PR #268, opened 2021-06-07
Use `built` to determine version information

djmitche:
The existing invocation of `git` wasn't working for me:
```
cleave ~/p/taskchampion [use-built] $ ta version                                                                                                                                                                                                                                                  
TaskChampion 0.3.0 (git rev: )                                                                                                                   
```

this PR does about the same thing, but outsourcing the version determination to Someone Else's Crate.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -536,6 +536,16 @@ dependencies = [
 ]
 
 [[package]]
+name = "built"
+version = "0.5.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "4f346b6890a0dfa7266974910e7df2d5088120dd54721b9b0e5aae1ae5e05715"
+dependencies = [
+ "cargo-lock",
+ "git2",
+]
+
+[[package]]
 name = "bumpalo"
 version = "3.6.1"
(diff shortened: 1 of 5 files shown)


dbr:
The "with un-committed changes" is a nice addition!

```
$ cargo run -- version
[...]
TaskChampion 0.3.0 (git version: v0.3.0-134-g6c9ba48)
```

I didn't really like the `git describe --tags --always` format as it seemed a bit confusing to show the tag on a revision which wasn't a tag (previous `version` output just showed the shortened commit hash) - but on reading https://git-scm.com/docs/git-describe#_examples

> i.e. the current head of my "parent" branch is based on v1.0.4, but since it has a few commits on top of that, describe has added the number of additional commits ("14") and an abbreviated object name for the commit itself ("2414721") at the end.

..it makes some sense - shows the "previous tag, and number of commits since it, and the short-hash" - useful to know "running something like v0.3.0 but with a lot of additional commits" versus "v0.3.0 with 1 additional commit"

dbr (approved):

djmitche:
I originally included the commit hash, too, but as you've noted it's already embedded in the git-describe output after the `g` character.

PR #273, opened 2021-06-13
Create a new doc if necessary in `ta config set`

### LINKED ISSUE #269: "ta config set" fails with no existing config
```
$ cargo run config set modification_count_prompt 3
[...]
Other(Could not read existing configuration file

Caused by:
    No such file or directory (os error 2))
```

(also the `Other(...`  is missing a trailing paren?)

djmitche:
Fixes #269.

### DIFF
--- a/cli/src/settings/settings.rs
+++ b/cli/src/settings/settings.rs
@@ -184,10 +184,16 @@ impl Settings {
                 .ok_or_else(|| anyhow!("Could not determine config file name"))?
         };
 
-        let mut document = fs::read_to_string(filename.clone())
-            .context("Could not read existing configuration file")?
-            .parse::<Document>()
-            .context("Could not parse existing configuration file")?;
+        let exists = filename.exists();
+
+        let mut document = if exists {
+            fs::read_to_string(filename.clone())
+                .context("Could not read existing configuration file")?
+                .parse::<Document>()
+                .context("Could not parse existing configuration file")?
+        } else {
+            Document::new()
+        };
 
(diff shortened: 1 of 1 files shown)


cli/src/settings/settings.rs
```diff
@@ -184,10 +184,34 @@ impl Settings {
                 .ok_or_else(|| anyhow!("Could not determine config file name"))?
         };
 
-        let mut document = fs::read_to_string(filename.clone())
-            .context("Could not read existing configuration file")?
-            .parse::<Document>()
-            .context("Could not parse existing configuration file")?;
+        // check if the file exists
+        let exists = match fs::metadata(&filename) {
```
dbr, line 188: I think this could be simplified to [`filename.exists()`](https://doc.rust-lang.org/std/path/struct.PathBuf.html#method.exists), unless there's a benefit to returning the errors at this point? (I suspect that any error that would occur here would be reported by one of the later calls anyway)

dbr (commented):
Good to merge as-is, although might be worth shortening the `let exists = ...` code? (or maybe pull it out into a `try_exists` method with a note saying it's implementing the nightly-only feature so it can be replaced at some point in future?)

cli/src/settings/settings.rs
djmitche: Oh, thanks!  I was looking for an `exists()` in `std::fs`, not in `Path`.

dbr (approved):
LGTM!

PR #275, opened 2021-06-14
Fix arg parsing on Windows

### LINKED ISSUE #274: Build fails on Windows
As I don't have an access to *nix machine for about a month (long story), attempting to build on Windows:
```
Running `rustc --crate-name taskchampion_cli --edition=2018 cli\src\lib.rs 
...
error[E0433]: failed to resolve: could not find `unix` in `os`
  --> cli\src\lib.rs:34:14
   |
34 | use std::os::unix::ffi::OsStringExt;
   |              ^^^^ could not find `unix` in `os`

error[E0599]: no method named `into_vec` found for struct `OsString` in the current scope
  --> cli\src\lib.rs:66:42
   |
66 |         .map(|oss| String::from_utf8(oss.into_vec()))
   |                                          ^^^^^^^^ method not found in `OsString`

error: aborting due to 2 previous errors

Some errors have detailed explanations: E0433, E0599.
For more information about an error, try `rustc --explain E0433`.
error: could not compile `taskchampion-cli`
```
Is there an OS-agnostic replacement for [std::os::unix::ffi](https://doc.rust-lang.org/std/os/unix/ffi/index.html)?.. How important is building on a non-unix? 
@djmitche , do you aim for TC to be operational on Windows/MacOS?

djmitche:
Fixes #274

### DIFF
--- a/cli/src/lib.rs
+++ b/cli/src/lib.rs
@@ -31,8 +31,7 @@ For the public TaskChampion Rust API, see the `taskchampion` crate.
 
 */
 
-use std::os::unix::ffi::OsStringExt;
-use std::string::FromUtf8Error;
+use std::ffi::OsString;
 
 // NOTE: it's important that this 'mod' comes first so that the macros can be used in other modules
 mod macros;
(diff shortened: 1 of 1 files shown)


dbr:
Looks good to me!

I'd kind of be inclined just to use `std::env::args()` which panics on invalid-UTf-8, as this error as it seems like it'd be very uncommon.. but given it's only a couple of lines in main, and will produce a slightly better error, this change seems good I think

dbr (approved):

PR #276, opened 2021-06-15
CI tests on Windows+macOS

dbr:
..hopefully. As per https://github.com/taskchampion/taskchampion/issues/274


dbr:
These failures now seem like legitimate failures which should be fixed by #275

djmitche (approved):
Looks good!  I merged from main, so if the checks are green, click merge :)

dbr:
🚀

PR #279, opened 2021-06-20
Changelog

dbr:
I figured we should probably have a `CHANGELOG` file of some kind - the releases section is fine, but:
1. It's nice having it recorded in the main repo also, not just on Github
2. It's good to have the original author write the changelog entry at the time of PR (even if they get edited or rewritten at release time)
3. Should make it easier at release time than retroactively describing all the changes based on `git log`

Also to avoid having every concurrent PR ending up with minor merge-conflict on the CHANGELOG.md file, I shoved together a simple version of [something like this](https://about.gitlab.com/blog/2018/07/03/solving-gitlabs-changelog-conflict-crisis/)

> Also simple process/script to manage entries in PR's without merge conflicts:
> 
> 1. Unreleased entries area stored as text files in ".changelogs/"
> 2. On release these are concatenated together and put in CHANGELOG.md
> 
> Script basically just add line of text to a "YYYY-MM-DD-branch.txt" formatted file

For example:
```
$ ./scripts/changelog.py add "Improved thing for reason"
$ ls .changelogs/
2021-06-20-changelog.txt
$ cat .changelogs/2021-06-20-changelog.txt
- Improved thing for reason
$ ./scripts/changelog.py build
## x.y.z - 2021-06-20
- Improved thing for reason
```

(it doesn't really *need* a script, a simple `echo ... > .changelogs/1999-01-01.txt` would work fine, it's mostly just to save me having to remember the current date 🙄)

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -7,6 +7,10 @@ It also means that things are changing quickly, and lots of stuff is planned tha
 If you would like to work on TaskChampion, please contact the developers (via the issue tracker) before spending a lot of time working on a pull request.
 Doing so may save you some wasted time and frustration!
 
+A good starting point might be one of the issues tagged with ["good first issue"][first].
+
+[first]: https://github.com/taskchampion/taskchampion/issues?q=is%3Aissue+is%3Aopen+label%3A%22good+first+issue%22
+
 # Other Ways To Help
 
 The best way to help this project to grow is to help spread awareness of it.
@@ -15,7 +19,6 @@ Tell your friends, post to social media, blog about it -- whatever works best!
 Other ideas;
  * Improve the documentation where it's unclear or lacking some information
  * Build and maintain tools that integrate with TaskChampion
- * Devise a nice TaskChampion logo
 
 # Development Guide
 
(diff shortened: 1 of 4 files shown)


djmitche (commented):
Looks good!  Can you add

 * Some guidance on how Changelog entries should be written, in CONTRIBUTING.md.   For Taskcluster we used [this](https://github.com/taskcluster/taskcluster/blob/main/dev-docs/best-practices/changelog.md), but the case here is a bit simpler.  I think only the "Snippet Text" but of that is important here, so that all of the changelog entries scan nicely when added to CHANGELOG.md.
 * Add instructions to RELEASING.md on how to handle this on release
 * Maybe add a GH Action that checks for a `.changelog/*` change and fails if none exists?

scripts/changelog.py
```diff
@@ -0,0 +1,64 @@
...
+    return changefiles
+
+def cmd_add(args):
+    text = args.text.strip()
+    if not text.startswith("- "):
+        text = "- %s" % text
+
+    timestamp = ymd()
+    branchname = git_current_branch()
+    fname = os.path.join(get_dir(), "%s-%s.txt" % (timestamp, branchname))
```
djmitche, line 37:
```suggestion
    fname = os.path.join(get_dir(), "%s-%s.md" % (timestamp, branchname))
```
so that editors get the syntax right..

CONTRIBUTING.md
```diff
@@ -44,8 +47,19 @@ You may be able to limit the scope of what you need to understand to just one cr
...
+We expect contributors to follow the [GitHub Flow](https://guides.github.com/introduction/flow/).
+Aside from that, we have no particular requirements on pull requests.
+Make your patch, double-check that it's complete (tests? docs? documentation comments?), and make a new pull request.
+
+Any non-trivial change (particularly those that change the behaviour of the application, or change the API) should be noted in the projects changelog.
+In order to manage this, changelog entries are stored as text files in the `.changelog/` directory at the repository root.
+
+To add a new changelog entry, you can simply run `python3 ./script/changelog.py add "Fixed thingo to increase zorbloxification [Issue #2](http://example.com)`
+
+This creates a file named `./changelogs/yyyy-mm-dd-branchname.txt` (timestamp, current git branch) which contains a markdown snippet.
```
djmitche, line 61:
```suggestion
This creates a file named `./changelogs/yyyy-mm-dd-branchname.md` (timestamp, current git branch) which contains a markdown snippet.
```

scripts/changelog.py
```diff
@@ -0,0 +1,64 @@
...
+    here = os.path.dirname(os.path.abspath(__file__))
+    return os.path.join(
+        here,
+        "../.changelogs")
+
+def get_changefiles() -> List[str]:
+    changedir = get_dir()
+    changefiles = []
+    for f in os.listdir(changedir):
+        if f.endswith(".txt") and not f.startswith("."):
```
djmitche, line 25:
```suggestion
        if f.endswith(".md") and not f.startswith("."):
```

djmitche (approved):
Sorry I didn't spot the `.txt` thing before!

scripts/changelog.py
dbr: Ah, good point!

djmitche (approved):

PR #282, opened 2021-09-05
Fix clippy and only run it with the MSRV

### LINKED ISSUE #281: Fix new clippy warnings, pin clippy version
@dbr suggested pinning clippy to the msrv so that we don't get spurious failures.

djmitche:
Fixes #281.


dbr:
LGTM!

PR #289, opened 2021-09-28
Add support for snapshots

### LINKED ISSUE #23: Implement snapshot requests
As described in the development docs: server requests a snapshot from the client, client produces one

djmitche:
Fixes #23.

@dbr do you mind having a look at these docs to check that they capture your comments on #23?

I made a few changes:
 * snapshot requests have only two levels of urgency
 * snapshot requests are included in AddVersion respnses, since that is where the client is likely to be up-to-date
 * when getting a snapshot, no version is included -- the caller just gets the latest snapshot


dbr:
Yep looks good to me!

djmitche:
> error[E0463]: can't find crate for `serde_derive` which `built` depends on

let's hope that's intermittent..

dbr:
Hmm it seems like that might be caused by a bug in Github Actions' caching?
https://github.com/actions-rs/cargo/issues/111 -> https://github.com/actions/cache/issues/403

Seems like updating to `cache@v2` might fix it? (can't remember specifics but I did update another project to the newer version and it solved a bunch of other weirdness I was having)

djmitche:
OK, this implements the server side of things.  Still to do:
 * client side
 * sever configuration, so that parameters can be user-controlled
 * server-side garbage collection (deleting versions)
 * include this in integration tests (set the GC / snapshot parameters to low values and observe that things work/break as expected)

djmitche:
@dbr do you think you'll have a chance to take a look?

djmitche:
All of those notes were specific to commits, so I squashed them into the relevant commits..

PR #292, opened 2021-10-02
Switch to pretty_assertions

djmitche:
This is a big diff but it's very formulaic.  I basically added pretty_assertions to every file with `mod test {` in it, and then ran `cargo fix` to remove the unnecessary imports.


taskchampion/src/workingset.rs
```diff
@@ -69,6 +69,7 @@ impl WorkingSet {
 #[cfg(test)]
 mod test {
     use super::*;
+    use pretty_assertions::{assert_eq};
```
dbr, line 72: I'm surprised `cargo fmt` didn't remove the redundant `{ ... }` brackets?

dbr (approved):
LGTM repeated 47 times 😁

taskchampion/src/workingset.rs
djmitche: Good point. I'm also surprised that didn't cause a CI failure.  I'll sed that out and land.

dbr:
I wonder if the formatting checks are not running on the `#[test]` targets?

djmitche:
It appears we're not actually running `cargo fmt` in CI.  I'll make another PR :)

PR #293, opened 2021-10-02
add a cargo-fmt job to CI

djmitche:
(I also bumped checkout  to v2 on the blind assumption that's better)

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -13,7 +13,7 @@ jobs:
     name: "Clippy"
 
     steps:
-      - uses: actions/checkout@v1
+      - uses: actions/checkout@v2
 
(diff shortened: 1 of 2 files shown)


djmitche:
(I'll land without the second commit on a positive review)

I was hoping this would have annotations on the diff for errors, but at least it fails.

dbr (approved):
I initially thought it might be worth pinning to `toolchain: "1.54"` like with clippy.. but rustfmt changes drastically less often, so probably better to change if/when it becomes an issue

> I was hoping this would have annotations on the diff for errors, but at least it fails.

Yep I think this is good as is. The imgui-rs project has the annotations things for these (..I think?) but I've never found them of much use - just seeing the formatting check fails is always enough

..so, LGTM!

PR #301, opened 2021-10-16
Drop tindercrypt, document encryption

### LINKED ISSUE #129: Only derive encryption key once
As #25 is written right now, every encryption and decryption operation derives the key again, and since it sues PBKDF2, that's pretty slow.  That seems to be the way tindercrypt is written, but https://github.com/apyrgio/tindercrypt/issues/6 may change that.

### LINKED ISSUE #299: Document encryption format
In order for the sync interface to be an API, we'll need to document how data is encrypted on the client.

If the current approach is too specific to the library currently in use, we should change that (in a followup issue).

djmitche:
This fixes #299:
 * drops use of tindercrypt
 * keeps using the same crypto primitives as tindercrypt, but simplifies the metadata encoding dramatically (see crypto::Envelope)
 * documents the encryption
 * implements a "cryptor" that derives the key once, which will save a lot of time when applying several versions (fixes #129).

The new format is incompatible, so replicas before and after this change cannot communicate.

The biggest design question here is the wisdom of the key derivation, and in particular of basing its salt on the client_key.  The result is that everything for a particular client_key is encrypted with the same key, but I think that's OK.  And if doing so weakens PBKDF2 in any way, I'm not that concerned: PBKDF2 is really only meant to avoid brute-forcing its input given its output.

In fact, could we just use the SHA256 of the encryption secret + client key as the encryption key?  That would be faster.  Would it lose any advantage?

---

### TODO
* [x] Switch to ChaCha20Poly1305
* [x] Add an integration test with a fixed payload or alternative implementation
* [x] Use u8 for the version


dbr:
I also wonder if it's worth adding some kind of integration test for the encryption protcol, as currently all the tests are roundtrips so would happily keep working even if the wire protocol was to accidentally change in some drastic way

First thing that springs to mind is a very basic implementation in a Python script or similar, which the test can run and exchange messages to/from

djmitche:
That's a great idea.  In fact, I could just use some fixed data -- no need to bring a second implementation into play.

djmitche:
(I'll wait for @savchenko to chime in re the crypto)

savchenko:
> I can't really comment on the crypto side of things beyond "sounds reasonable" - maybe @savchenko can have a quick look as he's far more familiar with that sort of thing

@savchenko wrote his first "hello world" in Rust yesterday (to which @dbr was a virtual witness).
After the cursory reading, minus all the Rust expressions I don't understand:

```rust
let salt = digest::digest(&digest::SHA256, client_key.as_bytes());
```

What is the `client_key`? If the value is computed in deterministic manner and reused, it might introduce the possibility of creating a rainbow table. I wouldn't stress too much about it given the other primitives used, but it's a good practice to guarantee a random salt.

`AES_256_GCM`

Given the application and its expected traffic, Chacha20/Poly1305 might be better suited.

_(Andrew notices the top message)_

> could we just use the SHA256 of the encryption secret + client key as the encryption key?

Err... Not quite following. Are you suggesting `sha256(client_key)` to encrypt the message? Same question as above, what is the `client_key` and how is it stored/generated? 

For those living on Mars (_/me puts on Martian costume_), is there an up-to-date text that describes desirable security properties of the Taskchampion as the whole system? Alternatively, I will be in Signal later today if you prefer to have a call.

djmitche:
> however I could very well be wrong

Now, you're spot-on.  If you squint just right, client_key is the username and encryption_secret is the password.  The server knows the client_key (uses it as a DB key) but not the encryption secret.

>  is there an up-to-date text that describes desirable security properties of the Taskchampion as the whole system?

https://github.com/taskchampion/taskchampion/commit/636862f8c5606bb9969c6bbeea8182c9a2e3ccc4#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5L24
```
 * Reasonable privacy: user's task details not visible on server
```
So basically the server just holds blobs of bytes and cannot recover the user's tasks.  By "reasonable" I mean I've not included key rotation or non-repudiation or anything fancy like that.  #302 addresses re-documenting this.

---

To summarize the question:

Current (pre-PR) situation: key is derived using PBKDF2 from the encryption_secret and a random salt with 100,000 iterations, then used to encrypt payloads sent to the server and decrypt those payloads when received.

Proposed: key is derived using SHA256 of the encryption_secret and the client_key, and used the same way.

Question: Does this lose any important properties?

My understanding is that PBKDF2 is meant to resist brute-force attacks where, given the key, an attacker would like to derive the encryption_secret.  In this implementation, the key is an attacker's ultimate goal (if an attacker has the key, they can decrypt everything), so such a brute-force attack would not be useful.

Note that AEAD requires a key never be re-used with the same nonce.  TC uses a random 96-bit nonce.

---

Regarding the crypto algorithm, AEAD does support ChaCha20Poly1305.  I only chose AES_256_GCM as that's what tindercrypt used.

djmitche:
Oh and

> @savchenko wrote his first "hello world" in Rust yesterday

Welcome to the light side :)

savchenko:
> key is derived using PBKDF2 from the encryption_secret and a random salt with 100,000 iterations, then used to encrypt payloads sent to the server and decrypt those payloads when received

Good! If there is no hard dependency, you might want to use Argon as per RFC9106.

> Question: Does this lose any important properties?

I think it does, sha256 accelerates stupidly well even on commonplace hardware. It becomes and exercise of `sha256(secret+known_value)` which is plain dangerous unless number of rounds is drastically increased.

Modern GPUs can do >9.5K Mh/s (9500000000 hashes per second) of the above-mentioned `sha256(secret+salt)`. 

If anything, changing to a (much) slower hash will improve the standing quite a bit.

>  In this implementation, the key is an attacker's ultimate goal

Depends on the attacker's goals. Are you doing _verify-after-decrypt_ or _decrypt-after-verify_?

djmitche:
> What is that "material to brute-force", in this case?

I may be misunderstanding what you're saying, but I think the "material to brute-force" here is the encryption key.  I don't see why an attacker, in possession of an encryption key, would bother to brute-force the value from which that key was derived.

djmitche:
Ah, reading https://en.wikipedia.org/wiki/Key_derivation_function#Key_stretching_and_key_strengthening I see why a slow function is used.  It's not to avoid brute-forcing the password given the key.  It's to avoid an offline brute-force attack that would generate the key by brute-forcing the much-smaller space of passwords.

What I want to avoid in TC is having to run a slow KDF for each version downloaded, as at 1s each for 100 versions that can be more than a minute to run `ta sync`.

So the options are
 * A faster KDF, so that it's easy to run it for each version
 * Use the same key for all versions, and just derive it once using a slow KDF
    * This has the complicating factor of where to store the KDF salt
 * Use the same key for all versions, and derive it on-demand using a simple KDF with no salt

The last option really only makes sense if we can trust users to use large, random secrets.  The docs say to generate a hex string, which helps -- but people are still going to use `hunter2`, aren't they?

So perhaps the second-to-last option is a good choice.  And how about using the clientId as the salt?  It's a fixed salt, but different for every user, so at least it's resistant to dictionary attacks.

WDYT?

PR #305, opened 2021-10-21
Client initialization from snapshots

djmitche:
This adds support for the `sync` operation to look for a snapshot if it's syncing an empty taskdb.

It also adds a specific error message for replica being out of sync, and handles that with a more verbose note in the cli.  This bit is kind of ugly, because (a) error handling in the taskchampion crate isn't great, and there's an issue for that; and (b) the cli output is kind of lame and not something we need to work on right now.


dbr:
Looks good to me!

Issue #312, opened 2021-10-27
Change `tag.foo` key to `tag_foo`


djmitche:
This is an accidental inconsistency in the TC and TW data models.

PR #313, opened 2021-10-30
Add support for annotations

djmitche:
This matches the taskwarrior task model for annotations.

Note that I didn't support `ta denotate`.  If we feel that's necessary, we can make an issue to do it in a followup (outside of 0.5.0).

### DIFF
--- a/cli/src/argparse/subcommand.rs
+++ b/cli/src/argparse/subcommand.rs
@@ -207,6 +207,13 @@ impl Modify {
                 "start" => modification.active = Some(true),
                 "stop" => modification.active = Some(false),
                 "done" => modification.status = Some(Status::Completed),
+                "annotate" => {
+                    // what would be parsed as a description is, here, used as the annotation
+                    if let DescriptionMod::Set(s) = modification.description {
+                        modification.description = DescriptionMod::None;
+                        modification.annotate = Some(s);
+                    }
+                }
                 _ => {}
             }
 
@@ -225,6 +232,7 @@ impl Modify {
                     arg_matching(literal("start")),
                     arg_matching(literal("stop")),
                     arg_matching(literal("done")),
+                    arg_matching(literal("annotate")),
                 )),
                 Modification::parse,
             )),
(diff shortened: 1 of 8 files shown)


cli/src/argparse/subcommand.rs
```diff
@@ -278,6 +286,13 @@ impl Modify {
                 Mark all tasks matching the required filter as completed, additionally applying any given
                 modifications.",
         });
+        u.subcommands.push(usage::Subcommand {
+            name: "annotate",
+            syntax: "<filter> annotate [modification]",
+            summary: "Mark tasks as completed",
```
dbr, line 292: Copypasto?

taskchampion/src/task/task.rs
```diff
@@ -145,6 +145,22 @@ impl Task {
...
+    pub fn get_annotations(&self) -> impl Iterator<Item = Annotation> + '_ {
+        self.taskmap.iter().filter_map(|(k, v)| {
+            if let Some(ts) = k.strip_prefix("annotation_") {
+                if let Ok(ts) = ts.parse::<i64>() {
+                    return Some(Annotation {
+                        entry: Utc.timestamp(ts, 0),
+                        description: v.to_owned(),
+                    });
+                }
+                // note that invalid "annotation_*" are ignored
```
dbr, line 158: Might be worth shoving a warning in `log`? (even as `log::debug!(...)`?)

taskchampion/src/task/task.rs
```diff
@@ -225,6 +241,20 @@ impl<'r> TaskMut<'r> {
         self.set_string(format!("tag.{}", tag), None)
     }
 
+    /// Add a new annotation.  Note that annotations with the same entry time
+    /// will overwrite one another.
```
dbr, line 245: This seems a little odd, but, presumably just constrained by TW?

dbr (changes requested):

taskchampion/src/task/task.rs
djmitche: Yeah, and in practice it's hard to make two annotations in one second :)

djmitche:
rebased on top of #312

taskchampion/src/task/task.rs
dbr: Only vaguely plausible thing I can think of is `ta 1 annotate Did thing; ta +active annotate Another thing in batch` but seems quite esoteric, doubly so if that's how TW has worked for hwoever long!

dbr (approved):
Seems good to me!

PR #316, opened 2021-12-18
Define UDAs

djmitche:
This defines UDAs and provides a basic Replica/Task API for setting, removing, getting, and enumerating them.

The docs update suggests UDAs with a namespaced format.  The idea is to avoid name collisions, either with future additions to the data model (by virtue of using `.`, while the data model separates things with `_`) or between applications using the API.  It's worth noting that no existing TW tools follow this format, but a programmer can dream.


dbr:
Seems good to me!

> It's worth noting that no existing TW tools follow this format, but a programmer can dream.

I wonder if the dream can be nudged closer to reality by making the `set_uda` method similar to

```rust
fn set_uda(&self, namespace, key, value) {
    let key = format!("{}.{}", namespace, key);
    ...
}
fn set_uda_no_namespace_because_i_am_lazy_developer(&self, key, value) {
    self.set_uda("", key, value);
}
```
or something maybe slightly less overtly judgemental.. but.. something to nudge API-users towards using namespaces

Idea being that it's nicer to write

```rust
static UDA_NS: &str = "example";

...

task.set_uda(UDA_NS, "githubid", id);
```
than
```rust
task.set_uda(format("{}.{}", UDA_NS, github_id), id);
```

And is also nicer than writing:
```rust
task.set_uda_no_namespace(github_id, id);
```

djmitche:
That's a great idea, thanks!

I appreciate your thoughtful reviews :)

djmitche:
OK, ready for another look now.

dbr (approved):
LGTM!

PR #319, opened 2021-12-20
Fix application of modifications during 'ta add'

### LINKED ISSUE #317: `ta add` doesn't accept tags
```
ta add foo +a +b
```
adds a task with decription "foo", but it doesn't have any tags.

djmitche:
Fixes #317.


dbr:
LGTM!

PR #320, opened 2021-12-20
use strum_macros::Display to display Status

### LINKED ISSUE #318: match values of `status` to those in TaskWarrior
TW uses `pending`, `completed', etc.  TC is using single-letter abbreviations.

djmitche:
Fixes #318.

### DIFF
--- a/taskchampion/src/replica.rs
+++ b/taskchampion/src/replica.rs
@@ -274,7 +274,7 @@ mod tests {
                     uuid: t.get_uuid(),
                     property: "status".into(),
                     old_value: None,
-                    value: Some("P".into()),
+                    value: Some("pending".into()),
                     timestamp: now,
                 },
                 ReplicaOp::Update {
@@ -294,8 +294,8 @@ mod tests {
                 ReplicaOp::Update {
                     uuid: t.get_uuid(),
                     property: "status".into(),
-                    old_value: Some("P".into()),
-                    value: Some("C".into()),
+                    old_value: Some("pending".into()),
+                    value: Some("completed".into()),
                     timestamp: now,
                 },
             ]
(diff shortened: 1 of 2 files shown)


dbr:
It seems accident prone to give access to out both `"Pending"` and `"pending"` in somewhat similar methods (Display impl and `to_taskmap`), especially given

```rust
assert_eq!(Status::from_taskmap("something-else"), Status::Pending);
```

..which presumably means `Status::from_taskmap("completed") != Status::from_taskmap("Completed")` (and one of those becomes `Pending`)

Possibly just `from_taskmap` should return a `Result::Err`, or does that cause other problems?

djmitche:
The idea here is that no taskdb can be "invalid" in the sense that the app completely fails to function, so `from_taskmap` should always succeed.  In practice, we're likely to see this on import from old taskwarrior DB's, where for example a status might be "waiting" (even though that status was dropped in TW 2.6.0).

Is there a specific case where you can see this accident occurring?

djmitche:
@dbr ^^ thoughts?

dbr:
> Is there a specific case where you can see this accident occurring?

The migration from other tools or older versions seems like the most likely - and also if people are using the API to write custom importers, it might be hard to spot that most ticket status translate correctly but the occasional imported `wontfix` status items end up `Pending`

Would it make sense to add some kind of `Status::Unknown` variant to handle these? (maybe even `Status::Unknown(String)`) - that way the parsing can always succeed, but still treat unexpected values distinct from actually-"Pending" ones

djmitche:
Yeah, that makes sense.  We might add more statuses eventually, so that would be a good forward-compatibility change.

djmitche:
@dbr, want to take another look?

dbr (approved):
LGTM!

PR #321, opened 2021-12-20
fix a few lints in clippy 1.57

djmitche:


dbr:
In light of the issue number, :rocket:

Issue #238, opened 2021-05-16
Pin the version of clippy


djmitche:
We keep getting failures on PRs for new clippy warnings unrelated to the PR.

Let's pin the version of clippy that's used in CI, so that this doesn't change unexpectedly.

djmitche:
Ah, this is coordinated with the Rust version, and indeed Rust 1.52 came out a week ago.  So I think this will be solved by #216.

PR #324, opened 2021-12-22
Update clippy toolchain to 1.57

djmitche:
It looks like we were just using an older clippy?

```
funk ~/p/taskchampion [issue322*] $ cargo clean 
funk ~/p/taskchampion [issue322*] $ cargo clippy 
   Compiling libc v0.2.97              
   Compiling proc-macro2 v1.0.27      
...
   Compiling built v0.5.1
    Checking replica-server-tests v0.4.1 (/home/dustin/p/taskchampion/replica-server-tests)
   Compiling taskchampion-cli v0.4.1 (/home/dustin/p/taskchampion/cli)
    Finished dev [unoptimized + debuginfo] target(s) in 1m 44s
funk ~/p/taskchampion [issue322] $ cargo version
cargo 1.57.0 (b2e52d7ca 2021-10-21)
```

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -31,7 +31,7 @@ jobs:
         with:
           # Fixed version for clippy lints.  Bump this as necesary.  It must not
           # be older than the MSRV in tests.yml.
-          toolchain: "1.54"
+          toolchain: "1.57"
           override: true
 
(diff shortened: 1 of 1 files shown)


dbr (approved):
Yep we pinned clippy to a specific version to stop spurious PR failures in #238 but it's worth updating every so often, LGTM

Issue #328, opened 2022-01-06
Add `ta import-tdb2` to import from TDB2 files


djmitche:
Let's support importing directly from TaskWarrior TDB2 files (`.task/*.data`).  This will allow us to avoid the data lost in `task export`, particularly type information for UDAs.

This will also represent a higher-fidelity test of the match between the TaskChampion and TaskWarrior data models.

PR #329, opened 2022-01-06
Implement `ta import` and `ta import-tdb2`

### LINKED ISSUE #94: Import TaskWarrior task list
To support transitioning from TW to TC, we'll want to be able to import the output of `task export`, at least capturing the basic set of tasks.  If that misses recurrences or some other not-very-comparable feature, that's probably OK.

djmitche:
Fixes #94.

Two caveats here:
 * UDAs are not imported correctly, since we lack type information.  My preferred fix is
   * #328
 * Several kinds of data are not yet supported in TaskChampion.  I've added TODO notes for those.


djmitche:
I implemented #328 on this branch, too.

I don't think it hurts to keep both import commands -- this is largely for testing anyway, right?

taskchampion/src/replica.rs
```diff
@@ -99,18 +102,25 @@ impl Replica {
...
         task.set_description(description)?;
         task.set_status(status)?;
+        trace!("task {} created", uuid);
         Ok(task.into_immut())
     }
 
+    /// Create a new, empty task with the given UUID.  This is useful for importing tasks, but
+    /// otherwise should be avoided in favor of `new_task`.  If the task already exists, this
+    /// does nothing and returns the existing task.
+    pub fn create_task(&mut self, uuid: Uuid) -> anyhow::Result<Task> {
```
dbr, line 118: Maybe this could have a clunkier name like `import_task_with_uuid` to differentiate it from the preferred `new_task`?

cli/src/argparse/subcommand.rs
```diff
@@ -424,6 +433,66 @@ impl Sync {
...
+            name: "import",
+            syntax: "import",
+            summary: "Import tasks",
+            description: "
+                Import tasks into this replica.
+
+                The tasks must be provided in the TaskWarrior JSON format on stdin.  If tasks
+                in the import already exist, they are 'merged'.
+
+                Because TaskChampion lacks the information about the types of UDAs that is stored
```
dbr, line 457: It feels a bit odd to have the `ta import` command be TW specific (rather than be a TC-specific "roundtripable" format), but, given the direction of things this is probably fine :shrug:

dbr: Maybe makes sense to have `ta import` be name `ta import-tw` or similar? Also ":shrug:"

djmitche:
Good points all, and implemented.

dbr (approved):
LGTM!

PR #330, opened 2022-01-06
Support an 'end' key in task maps

### LINKED ISSUE #327: Support `end` property for when a task is completed (or deleted?)
This is to match the TaskWarrior data model.

djmitche:
This definition matches how TaskWarrior uses the same key.

Fixes #327.

### DIFF
--- a/taskchampion/src/task/task.rs
+++ b/taskchampion/src/task/task.rs
@@ -265,7 +265,6 @@ impl<'r> TaskMut<'r> {
     /// Set the task's status.  This also adds the task to the working set if the
     /// new status puts it in that set.
     pub fn set_status(&mut self, status: Status) -> anyhow::Result<()> {
-        if status == Status::Pending {}
         match status {
             Status::Pending => {
                 // clear "end" when a task becomes "pending"
--- a/taskchampion/src/task/task.rs
+++ b/taskchampion/src/task/task.rs
@@ -57,6 +57,7 @@ enum Prop {
     Start,
     Status,
     Wait,
+    End,
 }
 
 #[allow(clippy::ptr_arg)]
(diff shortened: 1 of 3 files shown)


djmitche:
@tbabej can you confirm the docs update here?  I've interpreted it from the TW source but I might be wrong :)

These docs are sort of a "rough draft" for https://github.com/GothenburgBitFactory/tw.org/issues/212.

tbabej:
Looks correct to me!

taskchampion/src/task/task.rs
```diff
@@ -263,9 +264,23 @@ impl<'r> TaskMut<'r> {
     /// Set the task's status.  This also adds the task to the working set if the
     /// new status puts it in that set.
     pub fn set_status(&mut self, status: Status) -> anyhow::Result<()> {
-        if status == Status::Pending {
-            let uuid = self.uuid;
-            self.replica.add_to_working_set(uuid)?;
+        if status == Status::Pending {}
```
dbr, line 267: Unneeded line

taskchampion/src/task/task.rs
```diff
@@ -263,9 +264,23 @@ impl<'r> TaskMut<'r> {
...
+                // clear "end" when a task becomes "pending"
+                if self.taskmap.contains_key(Prop::End.as_ref()) {
+                    self.set_timestamp(Prop::End.as_ref(), None)?;
+                }
+                let uuid = self.uuid;
+                self.replica.add_to_working_set(uuid)?;
+            }
+            Status::Completed | Status::Deleted => {
+                // set "end" when a task is deleted or completed
+                if !self.taskmap.contains_key(Prop::End.as_ref()) {
```
dbr, line 279: Somewhat unrelated thing, but if `Prop` was `#[derive(Copy)]` would it neaten these checks to `if self.taskmap.contains_key(Prop::End)` ?

dbr (changes requested):

taskchampion/src/task/task.rs
djmitche: The `as_ref` here is returning `&'static str`, rather than `&Prop` (it's a strum method).  So implementing Copy doesn't allow removing `.as_ref()`.

dbr: Ahh, right!

PR #331, opened 2022-01-06
add 'entry' key to tasks when created

### LINKED ISSUE #326: Support `entry` property for when the task was created
This is to match TaskWarrior.

djmitche:
Fixes #326.


dbr:
`entry` for this is an odd choice of words (`created` or `created_at` or `insertion` etc would seem clearer to me), but I assume `entry` is necessary to match with TaskWarrior - if so, LGTM!

dbr (approved):

djmitche:
> `entry` for this is an odd choice of words (`created` or `created_at` or `insertion` etc would seem clearer to me), but I assume `entry` is necessary to match with TaskWarrior - if so, LGTM!

That's it exactly.

PR #336, opened 2022-02-22
Add support for dependencies

djmitche:
The TC API part of this is, so far, pretty simple.  Supporting modifications in the CLI was a little harder, as it requires "resolving" things like "depends:5" to a UUID, which requires a replica.  But that wasn't too bad, either.

The vritual tags support has me a little stumped, though.  +BLOCKED requires looking at the depended-on task to see if it's still pending.  That's not so bad, but requires a reference to the replica, which most `Task` methods do not (`TaskMut` does, but this isn't a mutation..).  +BLOCKING is even worse, though: it requires, I think, a full scan of the taskdb.

I'll look at how TaskWarrior implements these things.

### DIFF
--- a/taskchampion/src/task/task.rs
+++ b/taskchampion/src/task/task.rs
@@ -243,8 +243,25 @@ impl Task {
             .map(|(p, v)| (p.as_ref(), v.as_ref()))
     }
 
+    /// Get the modification time for this task.
     pub fn get_modified(&self) -> Option<DateTime<Utc>> {
         self.get_timestamp(Prop::Modified.as_ref())
+    }
+
+    /// Get the UUIDs of tasks on which this task depends.
+    ///
+    /// This includes all dependencies, regardless of their status.  In fact, it may include
+    /// dependencies that do not exist.
+    pub fn get_dependencies(&self) -> impl Iterator<Item = Uuid> + '_ {
+        self.taskmap.iter().filter_map(|(p, _)| {
+            if let Some(dep_str) = p.strip_prefix("dep_") {
+                if let Ok(u) = Uuid::parse_str(dep_str) {
+                    return Some(u);
+                }
+                // (un-parseable dep_.. properties are ignored)
+            }
+            None
+        })
     }
 
     // -- utility functions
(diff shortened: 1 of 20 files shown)


dbr:
> The vritual tags support has me a little stumped, though. +BLOCKED requires looking at the depended-on task to see if it's still pending. That's not so bad, but requires a reference to the replica, which most Task methods do not (TaskMut does, but this isn't a mutation..).

Does it make sense to move methods which do non-trivial things with inter-task references onto the `Replica` itself? E.g `Replica::is_blocked(&self, task: &Task) -> Something` or something - I guess it's akward as things like the virtual tags are in some sense as generic as "status: complete or not" but hides that there might be some complex query happening to determine the tag

> it requires, I think, a full scan of the taskdb

I'd be surprised if this is a practical problem - since I assume it is limited to active tasks (which could easily be optimized by SQL indexes in the storage layer), a full rescan is probably in the "tens of tasks" for most users, and even if it's thousands then still a full rescan probably isn't excessively slow

djmitche:
Good points, thanks!

In TaskWarrior, it's done in [`dependency_scan`](https://github.com/GothenburgBitFactory/taskwarrior/blob/e60fdafdaa461250d9a82ecfb20251473ba8c676/src/TDB2.cpp#L510-L543).  This appears to be called per-file, including for completed.data, but I'm not sure why that would be the case.

I think you're right that this will be fairly inexpensive if done infrequently.  I'd like to avoid repeating it in loops over tasks, though.  So, for example, let's not do this scan in `Replica.get_task`!  And, it may be useful (later?) to have a mode where this isn't calculated at all -- that might be useful in an integration like Bugwarrior, for example, which never reads tags.

So, maybe what I'll do is put this dependency information into a struct, similar to what's done with WorkingSet, and calculate that once in the replica before passing it to all created Tasks from that replica (so `Rc<RefCell<Dependencies>>` or something like that).

djmitche:
(rebased on top of the cdylib, and C API added)

Note that this doesn't add a `Task::get_dependents` function.  We can add that later.

taskchampion/src/task/task.rs
```diff
@@ -151,6 +165,16 @@ impl Task {
         self.taskmap.contains_key(Prop::Start.as_ref())
     }
 
+    /// Determine whether this task is blocked -- that is, has at least one unresolved dependency.
+    pub fn is_blocked(&self) -> bool {
+        self.depmap.dependencies(self.uuid).next().is_some()
```
dbr, line 170: Huh, kind of surprising the iterator doesn't have an equivelant of an `is_empty()` method (`.count() == 0` is the next closest thing, but `.next().is_some()` is probably more efficient)

dbr (approved):
LGTM!

taskchampion/src/task/task.rs
djmitche: Yeah, I went looking for such a thing.  The docs mention "empty iterator" lots of times, but no "is_empty".  I suspect that's because it's not possible to do non-destructively?

djmitche:
(rebased; I'll merge if this is green)

PR #339, opened 2022-03-06
use TC_UUID_STRING_BYTES constant directly

djmitche:
I spotted this while poking around elsewhere -- taskchampion-lib exposes a literal constant so that cbindgen can reflect its value into the header file.  But it was using a constant from the uuid crate in the code.  They had the same value, so it didn't make a big difference, but (a) this is cleaner and (b) this avoids taskchampion-lib requiring uuid directly.


dbr (approved):
LGTM!

PR #341, opened 2022-03-06
Add support for expiration

djmitche:
This expires _deleted_ tasks 180 days after their last modification.

In most cases, users' tasks will stay around forever, as they are marked "completed" and not "deleted", but users who wish to keep their task list short can do something like `ta +COMPLETED delete` periodically.

### DIFF
--- a/taskchampion/src/replica.rs
+++ b/taskchampion/src/replica.rs
@@ -4,7 +4,7 @@ use crate::task::{Status, Task};
 use crate::taskdb::TaskDb;
 use crate::workingset::WorkingSet;
 use anyhow::Context;
-use chrono::Utc;
+use chrono::{Duration, Utc};
 use log::trace;
 use std::collections::HashMap;
 use uuid::Uuid;
@@ -127,7 +127,6 @@ impl Replica {
     /// Delete a task.  The task must exist.  Note that this is different from setting status to
     /// Deleted; this is the final purge of the task.  This is not a public method as deletion
     /// should only occur through expiration.
-    #[allow(dead_code)]
     fn delete_task(&mut self, uuid: Uuid) -> anyhow::Result<()> {
         self.add_undo_point(false)?;
         self.taskdb.apply(SyncOp::Delete { uuid })?;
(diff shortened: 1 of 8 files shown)


taskchampion/src/task/task.rs
```diff
@@ -443,7 +443,9 @@ impl<'r> TaskMut<'r> {
         value: Option<String>,
     ) -> anyhow::Result<()> {
         let property = property.into();
-        self.lastmod()?;
+        if &property != "modified" {
+            self.lastmod()?;
```
dbr, line 447: Maybe worth renaming `lastmod` to `update_lastmod` or `set_lastmod`? Reads a little confusingly as sounds like it's an getter

taskchampion/src/replica.rs
```diff
@@ -392,4 +416,34 @@ mod tests {
         let uuid = Uuid::new_v4();
         assert_eq!(rep.get_task(uuid).unwrap(), None);
     }
+
+    #[test]
+    fn expire() {
```
dbr, line 421: Would it be worth adding a second test case to detail what happens with conflicts in this case - e.g replica 1 deletes task, a year later replica 2 marks task is done, another year later the two replicas are synced? (I assume it essentially creates a new, maybe incomplete, task with the modifications?)

taskchampion/src/task/task.rs
dbr: Also maybe worth a few more comments (e.g I assume the `if property != "modified"` check is to avoid recursion with lastmod being updated when setting last modified, but would probably be worth explicitly writing down)

taskchampion/src/replica.rs
djmitche: I believe (but can check) that is tested in detail by the sync tests, but it makes sense as an integration test.

taskchampion/src/task/task.rs
djmitche: Fixed in 393b804b30f062bd3fccae0199bfad55e4540161

taskchampion/src/replica.rs
djmitche: 47b0a8903325b64df99738de83b0b630ef543a92

dbr: Ah good point!

dbr (approved):

PR #345, opened 2022-03-12
Re-export the chrono crate from taskchampion.

djmitche:
The chrono types are central to use of TC, so this will help consumers
of the TC crate to avoid dependency conflicts.


dbr (approved):
Makes sense!

Issue #348, opened 2022-03-23
Support setting aribtrary k/v's in a task


djmitche:
Currently, the _only_ way to modify a task is to call a purpose-specific method like `set_description()` or `set_status()`.  This is a nice high-level interface, but people like low-level stuff, too.  In particular, TaskWarrior operates deeply at the k/v level, parsing or generating string values at the very "outside" of the application.  So, it will want to have something like `task.set_value(k, v)`.  This should be documented as a "low-level" interface.

miguelpduarte:
I can take this one, if that's okay! :smile:

djmitche:
all yours :D

djmitche:
@miguelpduarte how is this going?

miguelpduarte:
@djmitche I've been a bit slow to work on this as I'm still learning Rust and the code I'm touching seems to be a bit over my head. Looking at #350 helped, and I want to try to get a PR going (at least as a draft) in the following days.

I'll go ahead and ask a couple questions to try and speed this up as well:
* This issue mentions `set_description` and `set_status` as examples for functions that would be accessed via the to-be-implemented `set_value(k,v)`. From what I see these functions usually accept structs as arguments instead of strings. Is it expected that this method would handle the casting of values (possibly Strings) to the appropriate structs? Or is there any more adequate pattern that I should consider using?
* Furthermore, to get a general idea, is the aim of this method to "delegate" the set operations to the respective `set_x` methods, or is it also necessary to operate on the underlying task at times? I see that several of these methods operate using `self.set_string`. Would it make sense to wrap this method instead? It appears that it already expects a property that can be `Into`'d to a String, so it seems like a more direct approach. However, I'm not sure of the impact of one approach vs the other, especially in terms of checking for potentially invalid values for some properties, etc.
* Finally, in regards to testing, should I only consider using tests under `integration-tests`, or are there any unit tests on the Rust side? (or even, should they be added regardless of not existing yet)

PR #350, opened 2022-03-27
Treat priority as an arbitrary string

djmitche:
This matches what TaskWarrior does: priority is a UDA, and can be
redefined by the user's local config.

I dislike this decision, as it means that anything interpreting a task has to either know the user's UDA configurations (and at this point, configuration is not replicated, so that's a local setting); or guess that "L", "M", and "H" mean what they usually do.  But, it's how TW works so we'll stick with it.

@miguelpduarte this will touch similar bits of the code to #348, if you'd like to take a look.


dbr (approved):
Agreed - seems worse than having an inherently sortable type, and probably hard to change given TW's historical baggage, so, :shrug: LGTM

PR #354, opened 2022-04-23
export taskchampion-lib as an rlib, too

djmitche:
This puts all of the symbols at the top level, and publishes the crate as an rlib.

I suspect this will be the easiest way to make this C code available elsewhere?  At any rate, it was the simplest way I could find to integrate with Taskwarrior.

### DIFF
--- a/lib/Cargo.toml
+++ b/lib/Cargo.toml
@@ -4,7 +4,7 @@ version = "0.1.0"
 edition = "2018"
 
 [lib]
-name = "taskchampion"
+name = "taskchampionlib"
 crate-type = ["staticlib", "cdylib", "rlib"]
 
 [dependencies]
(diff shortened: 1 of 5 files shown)


dbr (approved):
LGTM!

dbr:
Barring the CI failures of course :grimacing:

lib/Cargo.toml
```diff
@@ -5,7 +5,7 @@ edition = "2018"
 
 [lib]
 name = "taskchampion"
-crate-type = ["staticlib", "cdylib"]
+crate-type = ["staticlib", "cdylib", "rlib"]
```
djmitche, line 8: I suspect this needs to be `taskchampion-lib` instead, to fix the lint issues.

PR #381, opened 2024-04-24
Batch SyncOp's to limit Version size.

### LINKED ISSUE #3331: sync protocol client should limit request body size
Given a huge set of un-sync'd operations, the sync protocol will generate a large version blob containing all of those operations. That might cause issues as in #3328, where a large request body was rejected.

Without any modification to the protocol, we could probably modify the client to break that set of operations up into smaller batches, and make a version for each one.  That might be a little complicated if some remote operations need to be included first, but maybe that will be handled with the existing mechanics.

ryneeverett:
This will effectively limit the size of http requests. Resolve https://github.com/GothenburgBitFactory/taskwarrior/issues/3331.

I hope this is what you had in mind. As always, please review with a heavy dose of skepticism.

### DIFF
--- a/taskchampion/src/server/test.rs
+++ b/taskchampion/src/server/test.rs
@@ -58,6 +58,11 @@ impl TestServer {
         let mut inner = self.0.lock().unwrap();
         inner.versions.remove(&parent_version_id);
     }
+
+    pub(crate) fn versions_len(&self) -> usize {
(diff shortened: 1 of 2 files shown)


djmitche (approved):
This looks great, and matches exactly what I had in mind -- thank you!

PR #382, opened 2024-04-24
Allow to access sync server with full URL instead of origin only

### LINKED ISSUE #367: Allow sync origins containing paths
From GothenburgBitFactory/taskwarrior#3361, we should allow the sync server to be configured by URL, not just origin. This will require
 * changing the name of the `sync.server.origin` config to `sync.server.url`, but keeping backward compatibility
 * parse it as a URL, and generate the paths for the various requests by appending to the configured path

This will involve some minor changes in `taskchampion/taskchampion/src/server/sync/mod.rs` as well as updating everywhere that `sync.server.origin` occurs to prefer `sync.server.url`.

dahelb:
resolves #367.

I renamed "origin" to "url" in `ServerConfig` and the C-bindings. For the `SyncServer`, I decided to use "base_url" instead to make clear that it is used to build the final endpoint URLs by appending to it.

I changed the type in `SyncServer` to be `Url` and added the new method `construct_endpoint_url` to build the final URLs.

### DIFF
--- a/docs/src/http.md
+++ b/docs/src/http.md
@@ -1,7 +1,7 @@
 # HTTP Representation
 
-The transactions in the sync protocol are realized for an HTTP server at `<origin>` using the HTTP requests and responses described here.
-The `origin` *should* be an HTTPS endpoint on general principle, but nothing in the functonality or security of the protocol depends on connection encryption.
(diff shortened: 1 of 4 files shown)


djmitche (approved):
Looks great -- nice work!

Apologies for the delay in reviewing..

Issue #372, opened 2023-02-04
Reconsider "level" of the Task/TaskMut interface


djmitche:
The problem that brought me [to GothenburgBitFactory/taskwarrior#3029] is, Taskwarrior _mostly_ doesn't call functions like `set_description` or `add_annotation` -- it interfaces directly with the key/value pairs that make up a task, via `set_value`.  But it does call `set_status` since that updates the working set.  So the issue was that Taskwarrior would call

 * `t.set_value("modified", "12345")`
 * `t.set_status(Status::Pending)`

and the second call would not "see" that the modified timestamp had already been set, and would do so again.  On import, this causes a problem by setting the "modified" timestamp of all tasks to the time the import occurred.

Overall, this is a major "impedance mismatch" between TC and TW: TW wants to handle tasks as key/value maps, while TC was really designed for API consumers to call methods that embody more knowledge about the meaning of particular keys.  These are more than just convenience wrappers, though: they also have this set-the-modified-timestamp thing, as well as automatically updating the working set.  And (referring to GothenburgBitFactory/taskwarrior#3028) it might be good to cache parent/child relationships as well, to avoid having to load most of the pending tasks one-by-one.  I've been mulling this over for a while but haven't come up with a concrete way to do better.

_Originally posted by @djmitche in https://github.com/GothenburgBitFactory/taskwarrior/issues/3029#issuecomment-1414569579_

djmitche:
I've gotten a start to this now, in #413. I'll leave some notes here that I will probably return to and update later.

## Notes

### General Goal

 * Tasks can be read and queried, with tasks represented as a simple wrapper around a HashMap.
 * Changes to tasks are represented as `Operation` instances, collected in an `Operations` type that is a thin wrapper around `Vec<Operation>`. New operations can be created in a variety of ways such as `Replica::new_task`, `Task::set_property`, or `Task::delete`, all of which take `&mut Operations`. (Note that this is, in essence, #373)
 * Operations are "committed" to the task DB by passing the `Operations` to a Replica method.
 * All of this blends nicely with the current undo support: the latest batch of undo operations are represented as `Operations`, and committing that undo involves reversing the operations and committing the result, with an optimization of "popping" matching operations from the DB's operations (as happens now).

### Complications

[...]

PR #388, opened 2024-05-13
WIP: Taskchampion bindings for python

illya-laifu:
# Taskchampion bindings for python

## Installation

The dependency management is done via `poetry`, to install the dependencies, from the  `py-lib` run:
```bash
poetry install
```
To build and install the package into current environment:
```bash
poetry run maturin develop
```
To run tests:
```bash
poetry run pytest
```

## Currently implemented and tested:
- `Annotation` class and the getters/setters for the two fields.
- `Replica` class (`::dependency_map` method needs more testing and polishing due to the `Rc` reference)
- `Status` enum
- `Tag` class
- `Task` class *
- `WorkingSet` class

## Implemented but not tested:
- `DependencyMap` class
- `Storage` concrete classes.

## TODO:
- Be able to construct replica by passing in the specific Storage implementations (e.g. `InMemoryStorage`, `SqliteStorage` or users own implementation of `Storage`)
- Current `Task` class is the immutable version of it, it would be nice to extend it to also include the methods from `TaskMut` and convert to and the mutable state in those method calls. It is impossible to implement the `TaskMut` via pyo3 due to the lifetime parameters.*
- Be able to iterate over WorkingSet via `next()` and with `for item in ws:`, but this needs it to hold the reference to an iterator.
- Python has it's own `uuid` library, currently the methods that accept a UUID, do that by accepting `str` and converting it into the rust's implementation of `UUID`. 
- Better documentation, but the Rust comments on the pyo3 annotated functions/objects directly translate into the docstrings.
- Not entirely sure if it is possible (need to research), but it would be convenient to provide signature stub files along with the library, so the Language Servers, linters and such would be able to infer the types.
- Either split this into a separate repo, since `maturin` generates multiple github workflows, or rename the `py-lib` into something sensible.


py-lib/README.md
```diff
@@ -0,0 +1,10 @@
+# Python Taskchampion Bindings
+
+This submodule contains bindings to the Taskchampion
+
+# TODO
+
+- There is no good way to describe functions that accept interface (e.g. `Replica::new` accepts any of the storage implementations, but Python bindings lack such mechanisms), currently, `Replica::new` just constructs the SqliteStorage from the params passed into the constructor.
+- Currently Task class is just a reflection of the rust's `Task` struct, but constructing the standalone `TaskMut` is impossible, as Pyo3 bindings do not allow lifetimes (python has no alternatives to them). Would be nice to expand the `Task` class to include the methods from `TaskMut` and convert into the mutable state and back when they are called.
+- It is possible to convert `WorkingSet` into a python iterator (you can iterate over it via `for item in <blah>:` or `next(<blah>)`), but that needs a way to store the current state.
```
djmitche, line 9: All three of these are valid criticisms of the Rust API, and problematic for embedding in other languages, too. I'd be happy to see PRs to simplify these Rust APIs. In fact, #372 is one such.

djmitche (commented):
I don't know Py03 well enough to really review this in detail, but structurally this looks great!

When this is ready for review, I will try to learn enough to verify
 * it does what it says on the tin
 * it does not make any promises to users that we cannot keep (e.g., making an assumption that TaskChampion doesn't actually support)
 * it seems safe :)

I have a minor concern that refactors of TC (notably #372) will cause breaking changes to bindings like this. However, I _think_ that's OK at this stage -- we can make a major version release of TC, with well-documented required changes, and consumers can upgrade as necessary.

PR #394, opened 2024-06-05
Fix credential path

### LINKED ISSUE #380: Fix redundant guards
`cargo clippy` will show a few errors about redundant guards in `taskchampion/src/server/sync/mod.rs`. Fix those up!

QuantumBJump:
A few things fixed here:
* `credential_path.unwrap() == ""` doesn't work - `unwrap()` forcibly crashes the program if the `Option` is `None`, therefore it can't be compared to a string. `is_none()` is what we want here.
* the `credential_path.is_none()` was impossible to evaluate as true anyway, because if the credential path wasn't set, the `unwrap()` on line 32 would have crashed the program before line 33 got to resolve.
* moving the setting of `credentialpathstring` after the `is_none()` check (which doesn't take ownership of the variable) both fixes that and removes the need for the `clone()`.

This should mean that if the default credential path isn't set, `ClientConfig::default().with_auth()` which was unreachable before, should successfully run.

### DIFF
--- a/taskchampion/src/server/cloud/gcp.rs
+++ b/taskchampion/src/server/cloud/gcp.rs
@@ -180,7 +180,7 @@ impl<'a> ObjectIterator<'a> {
     fn fetch_batch(&mut self) -> Result<()> {
         let mut page_token = None;
         if let Some(ref resp) = self.last_response {
-            page_token.clone_from(&resp.next_page_token.clone());
+            page_token.clone_from(&resp.next_page_token);
         }
         self.last_response = Some(self.service.rt.block_on(self.service.client.list_objects(
             &objects::list::ListObjectsRequest {
--- a/taskchampion/src/server/cloud/gcp.rs
+++ b/taskchampion/src/server/cloud/gcp.rs
@@ -29,12 +29,11 @@ impl GcpService {
     pub(in crate::server) fn new(bucket: String, credential_path: Option<String>) -> Result<Self> {
         let rt = Runtime::new()?;
 
-        let config: ClientConfig = if credential_path.is_none() {
-            rt.block_on(ClientConfig::default().with_auth())?
-        } else {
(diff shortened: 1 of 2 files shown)


taskchampion/src/server/cloud/gcp.rs
```diff
@@ -181,7 +180,7 @@ impl<'a> ObjectIterator<'a> {
     fn fetch_batch(&mut self) -> Result<()> {
         let mut page_token = None;
         if let Some(ref resp) = self.last_response {
-            page_token = resp.next_page_token.clone();
+            page_token.clone_from(&resp.next_page_token.clone());
```
djmitche:
```suggestion
            page_token.clone_from(&resp.next_page_token);
```

taskchampion/src/server/cloud/gcp.rs
```diff
@@ -29,12 +29,11 @@ impl GcpService {
...
-        let credentialpathstring = credential_path.clone().unwrap();
-        let config: ClientConfig = if credential_path.unwrap() == "" {
-            rt.block_on(ClientConfig::default().with_auth())?
-        } else {
-            let credentials = rt.block_on(CredentialsFile::new_from_file(credentialpathstring))?;
+        let config: ClientConfig = if let Some(credentials) = credential_path {
+            let credentials = rt.block_on(CredentialsFile::new_from_file(credentials))?;
             rt.block_on(ClientConfig::default().with_credentials(credentials))?
+        } else {
+            rt.block_on(ClientConfig::default().with_auth())?
```
djmitche, line 36: Yikes, that was on me for not reviewing carefully. Thanks for fixing it up!

djmitche (changes requested):
One little fix, if you agree. Thank you!

djmitche (approved):

PR #396, opened 2024-06-08
Overhaul documentation after TW merge

### LINKED ISSUE #392: Update TC docs to match reality
* There is no `ta` command anymore
* TC is a part of TW
* Document current status of tc-lib.

djmitche:
Fixes #392. I've tried to represent the current status accurately, without too much "future plants".

Some of this -- the task model, the sync model, and the HTTP spec -- are pretty critical for interoperation between GBF repos and with other projects, so I think it's important to get this published clearly.


docs/src/introduction.md
```diff
@@ -0,0 +1,11 @@
+# TaskChampion
+
+TaskChampion implements the task storage and synchronization behind Taskwarrior.
+It includes an implementation with Rust and C APIs, allowing any application to maintain and manipulate its own replica.
+It also includes a specification for tasks and how they are synchronized, inviting alternative implementations of replicas or task servers.
+
+## Relationship with Taskwarrior
+
+TaskChampion was originally developed as a project "inspired by" Taskwarrior, and later incorporated into Taskwarrior in its 3.0 release.
+Taskwarrior embeds TaskChampion, but does not have any kind of privileged access to its implementation details.
```
ryneeverett: I find this usage of the word "embed" a little odd.

```suggestion
Taskwarrior depends on TaskChampion, but does not have any kind of privileged access to its implementation details.
```

docs/src/introduction.md
```diff
@@ -0,0 +1,11 @@
...
+Any other application can also embed TaskChampion and implement similar functionality, and even interoperate with Taskwarrior either in the same replica or via sync.
```
ryneeverett: I find this usage less problematic because it more clearly means "embed" in the vague sense...but maybe an inherently vaguer word is better? :man_shrugging: 

```suggestion
Any other application can also use TaskChampion to implement similar functionality, and even interoperate with Taskwarrior either in the same replica or via sync.
```

ryneeverett (approved):

docs/src/introduction.md
djmitche: Thanks! I couldn't find the right word, and I like your choice better.

PR #397, opened 2024-06-10
Fix `get_child_version` returning versions not reachable from latest

### LINKED ISSUE #387: `get_child_version` must check that the version it returns is reachable from `latest`
This issue was identified in https://github.com/GothenburgBitFactory/taskwarrior/issues/3395. Here `get_child_version` is seeing that there is a single object in the bucket with the given parent version, and returning it. But, that version is not actually valid, as `latest` still points to the parent version.

djmitche:
When a previous sync attempt has uploaded a new child version of `latest`, but not updated `latest`, it sure looks like that version is a child of `latest` -- but `get_child_version` should never return a version not reachable from latest.

This includes a regression test for this scenario.  Fixes #387.

### DIFF
--- a/taskchampion/src/server/cloud/server.rs
+++ b/taskchampion/src/server/cloud/server.rs
@@ -392,14 +392,13 @@ impl<SVC: Service> Server for CloudServer<SVC> {
     }
 
     fn get_child_version(&mut self, parent_version_id: VersionId) -> Result<GetVersionResult> {
-        // The `get_child_versions` function will usually return only one child version for a
-        // parent, in which case the work is easy. Otherwise, if there are several possible
-        // children, only one of those will lead to `latest`, and importantly the others will not
-        // have their own children. So we can detect the "true" child as the one that is equal to
-        // "latest" or has children.
+        // The `get_child_versions` function may return several possible children, only one of
+        // those will lead to `latest`, and importantly the others will not have their own
+        // children. So we can detect the "true" child as the one that is equal to "latest" or has
+        // children. Note that even if `get_child_versions` returns a single version, that version
+        // may not be valid and the appropriate result may be NoSuchVersion.
         let version_id = match &(self.get_child_versions(&parent_version_id)?)[..] {
             [] => return Ok(GetVersionResult::NoSuchVersion),
-            [child] => *child,
             children => {
                 // There are some extra version objects, so a cleanup is warranted.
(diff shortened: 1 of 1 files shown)


taskchampion/src/server/cloud/server.rs
```diff
@@ -948,6 +948,29 @@ mod tests {
...
     }
 
+    #[test]
+    fn get_child_version_single_invalid() {
+        // This is a regression test for #387.
+        let mut server = make_server();
+        let (v1, v2, v3) = (Uuid::new_v4(), Uuid::new_v4(), Uuid::new_v4());
+        // Here v2 is latest, so v1 is not a valid child of v2.
+        server.mock_add_version(v2, v1, 1000, b"first");
+        server.mock_add_version(v3, v2, 1000, b"second");
```
ryneeverett:
```suggestion
        server.mock_add_version(v1, v2, 1000, b"second");
        server.mock_add_version(v2, v3, 1000, b"third");
```
If I understand this correctly, this ordering and naming makes more sense:
- The fourth `data` argument is associated with the second `child` argument, right?
- Children are newer than their parents, right? So (as the first assertion demonstrates) the first version is `v3` and the last version is `v1`. That seems confusing.

taskchampion/src/server/cloud/server.rs
```diff
@@ -948,6 +948,29 @@ mod tests {
...
+        server.mock_set_latest(v2);
+        assert_eq!(
+            server.get_child_version(v3).unwrap(),
+            GetVersionResult::Version {
+                version_id: v2,
+                parent_version_id: v3,
```
ryneeverett:
```suggestion
            server.get_child_version(v1).unwrap(),
            GetVersionResult::Version {
                version_id: v2,
                parent_version_id: v1,
```

taskchampion/src/server/cloud/server.rs
djmitche: Huh, I thought all of the tests had this backward ordering, but now that I look again that's not the case at all.

ryneeverett (approved):

taskchampion/src/server/cloud/server.rs
ryneeverett: That sounds familiar. I think maybe it was like that until i complained. :)

PR #398, opened 2024-06-10
Include docs for GCP access

### LINKED ISSUE #386: Add documentation for cloud config
There's some documentation in TaskWarrior's `task-sync(5)`, but that information should also be here in the API docs for ServerConfig.

djmitche:
Include better documentation for the parameters to `ServerConfig::Gcp`, based on those in Taskwarrior's task-sync(5) manpage.

Where the manpage is written as a guide for someone with minimal GCP familiarity, this documentation refers to GCP documentation instead, but includes the unique information required: the list of permissions.

Fixes #386.

### DIFF
--- a/taskchampion/src/server/config.rs
+++ b/taskchampion/src/server/config.rs
@@ -31,17 +31,35 @@ pub enum ServerConfig {
         /// be any suitably un-guessable string of bytes.
         encryption_secret: Vec<u8>,
     },
-    /// A remote taskchampion-sync-server instance
+    /// A Google Cloud Platform storage bucket.
     #[cfg(feature = "server-gcp")]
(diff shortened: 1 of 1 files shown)


ryneeverett (approved):
I don't understand why the audience for this is different from that of the taskwarrior manpage. But regardless, this is an improvement.

If a substantial amount of duplication emerges between taskchampion and taskwarrior docs, I doubt it would be too difficult to "generate" the taskwarrior manpages by inserting taskchampion docs into a template. But I don't think it's a big enough problem yet.

djmitche:
Thanks! If the distinction in audience becomes important, we can sort it out at that time.

PR #402, opened 2024-06-16
Include HTTP status text in error messages from ureq

djmitche:
The default display of a ureq::Error (which ends up being used by anyhow::Error) just includes the HTTP status, not the status text. So, this overrides the conversion to include the status text as well.

This also presents errors from ureq as Error::Server, rather than Error::Other.

This addresses GothenburgBitFactory/taskwarrior#3332.

### DIFF
--- a/taskchampion/src/errors.rs
+++ b/taskchampion/src/errors.rs
@@ -34,8 +34,6 @@ macro_rules! other_error {
         }
     };
 }
-#[cfg(feature = "server-sync")]
-other_error!(ureq::Error);
 other_error!(io::Error);
 other_error!(serde_json::Error);
 other_error!(rusqlite::Error);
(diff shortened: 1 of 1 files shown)


taskchampion/src/errors.rs
```diff
@@ -45,4 +43,41 @@ other_error!(google_cloud_storage::http::Error);
...
+            ureq::Error::Status(status, response) => {
+                let msg = format!(
+                    "{} responded with {} {}",
+                    response.get_url(),
+                    status,
+                    response.status_text()
+                );
+                Self::Server(msg)
+            }
+            ureq::Error::Transport(_) => Self::Server(ureq_err.to_string()),
```
ryneeverett, line 60: It might be nice to add a test for the `Transport` case as well since we're introducing the possibility of a mistake here and these errors might be so unusual and non-deterministic that a mishandling would result in difficult-to-diagnose "phantom bugs".

ryneeverett (approved):

taskchampion/src/errors.rs
djmitche: True, but quite a bit harder as those are not constructible (no new method)

PR #393, opened 2024-06-04
Fix cargo config file

QuantumBJump:
since cargo v1.38, `.cargo/config` has been deprecated in favour of `config.toml`. This updates the config file to stop later versions of cargo from complaining, while also putting a symlink at `.cargo/config` to ensure continued compatibility with versions < 1.38.

PR #405, opened 2024-06-21
Make `Replica::delete_task` public

### LINKED ISSUE #383: Add support for purging tasks
..meaning actually removing them from the task DB. This already occurs when tasks expire (#341), but sometimes (https://github.com/GothenburgBitFactory/taskwarrior/issues/3399) users want to purge a task without waiting for expiration.

It's OK for this to break the sync model a little bit, as long as that's clear in the documentation.  #26 has some discussion of this.

djmitche:
This has suitable warnings about how it interacts with sync. Note that the interactions don't actually _break_ sync: all replicas will still end up with the same task data. It just may not be what a user would expect. Hopefully the example in the comment makes that clear.

This also fixes the symlink in `.cargo` to have the correct name (`config.toml`) and link to the right place (`../config.toml`). This was missed in #393.


djmitche:
cc @QuantumBJump for the symlink fix :)

ryneeverett (approved):
I agree that the warnings are sufficient, but would it make sense to consider renaming these methods to `purge_task` to avoid reusing the term "delete" to mean two different things?

djmitche:
> I agree that the warnings are sufficient, but would it make sense to consider renaming these methods to `purge_task` to avoid reusing the term "delete" to mean two different things?

It ends up recording a Delete operation so in that sense the name makes sense. Also, purge has a connotation of removing all trace, yet the task data remains in the operation history. 

Naming things is hard :(

PR #409, opened 2024-06-29
Remove taskchampion-lib

djmitche:
This was originally intended to be a general-purpose C API for TaskChampion, but it's really not great. It's also quite difficult to use as a crate, rather than including the code within the emebedding project, as Taskwarrior has done. With this change, TaskChampion is just a Rust library, used by Taskwarrior but usable by anything.

The integration tests mostly tested the C API, and those are removed. However, two provided integratoin tests for the taskchampion crate, and those are moved to `taskchampion/tests`.

This was prompted by the necessity of duplicating changes to `lib` both here and in the copy of the same code in https://github.com/GothenburgBitFactory/taskwarrior.


djmitche:
(I jumped the gun a little and [marked the crate as deprecated](https://docs.rs/taskchampion-lib/latest/taskchampion_lib/). Sorry!)

ryneeverett:
The CI codegen failure would seem to be indicative.

ryneeverett (approved):
Makes sense to me. There's not much sense in promising the stability of this API if we're not committed to the implementation and taskwarrior can't even dogfood it.

PR #411, opened 2024-07-04
Replace "garbage-collection" with more precise words

djmitche:
I was sloppy in how I used this phrase!

See https://github.com/GothenburgBitFactory/tw.org/issues/877, cc: @sebastiancarlos.


sebastiancarlos (approved):
LGTM

PR #412, opened 2024-07-06
Use SQLite in WAL mode, with IMMEDIATE transactions

### LINKED ISSUE #3325: Racing working-set updates cause "Clear working set query" errors
Details in #3323.

A good place to start here would be to reproduce this issue: create a task db with a bunch of tasks in it and then run a loop displaying `task` output in such a way that it will clear and rebuild the working set every time. It will probably be necessary to run several such loops at the same time (e.g., in different terminal windows).

Once it's reproduced, the next step is to figure out what the actual error is, and maybe why that's not being shown.

Ideally Taskwarrior's DB should be resilient to concurrent access, but perhaps this is a transaction-failure error that occurs when one process writes entries into the working set concurrently with another process executing `delete from working_set`. If so, perhaps we can catch and re-try that transaction.

djmitche:
WAL mode is generally better, and allows concurrent reads to happen concurrently with other reads and writes. It does not allow writes concurrently with writes, though, failing with SQLITE_BUSY if that occurs.

SQLite's `busy_timeout` allows retrying on SQLITE_BUSY, but that does not work if a read-only transaction is upgraded to a write transaction, which is what `TransactionBehavior::Deferred` (a.k.a. `BEGIN DEFERRED`) does. `TransactionBehavior::Immediate` always begins a write transaction immediately, fails with SQLITE_BUSY if that is not allowed, and uses the `busy_timeout` logic to try to wait for the running transaction to finish.

This includes a unit test, but as the issue is a race condition, the test does not always fail without the fix. However, it reliably passes with the fix, which is really the important bit.

This fixes GothenburgBitFactory/taskwarrior#3325.

### DIFF
--- a/taskchampion/src/storage/sqlite.rs
+++ b/taskchampion/src/storage/sqlite.rs
@@ -2,7 +2,7 @@ use crate::errors::Result;
 use crate::storage::{ReplicaOp, Storage, StorageTxn, TaskMap, VersionId, DEFAULT_BASE_VERSION};
 use anyhow::Context;
 use rusqlite::types::{FromSql, ToSql};
-use rusqlite::{params, Connection, OpenFlags, OptionalExtension};
+use rusqlite::{params, Connection, OpenFlags, OptionalExtension, TransactionBehavior};
 use std::path::Path;
 use uuid::Uuid;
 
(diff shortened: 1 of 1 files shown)


taskchampion/src/storage/sqlite.rs
```diff
@@ -816,4 +823,32 @@ mod test {
...
+            // First thread begins a transaction, writes immediately, waits 100ms, and commits it.
+            scope.spawn(|| {
+                let mut storage = SqliteStorage::new(tmp_dir.path(), true).unwrap();
+                let u = Uuid::new_v4();
+                let mut txn = storage.txn().unwrap();
+                txn.set_base_version(u).unwrap();
+                thread::sleep(Duration::from_millis(100));
+                txn.commit().unwrap();
+            });
+            // Second thread begins a transaction, waits 50ms, and begins a transaction. This
```
ryneeverett:
```suggestion
            // Second thread waits 50ms, and begins a transaction. This
```

ryneeverett (approved):

PR #413, opened 2024-07-07
Refactor replica operations to a top-level type in the API

djmitche:
This is the beginning of #372, and makes `Operation` (formerly `ReplicaOp`) a top-level type that can be part of the TC API.

ReplicaOp had a few methods related to `SyncOp`, which is not a public type, so these were not useful in the public API. Those have been removed. However, the `taskchampion::storage::ReplicaOp` type still exists (now as a type alias) to try to minimize changes to the API. I suspect #372 will be a breaking change, anyway, so this is probably superfluous.


ryneeverett (approved):
Looks good. The approach you've outlined in #372 seems quite promising and I look forward to it being proven in subsequent PRs.

djmitche:
Thanks!

Issue #415, opened 2024-07-07
Replace `set!` with `HashSet::from`


djmitche:
The comment at the top was written when we were still compatible with Rust < 1.56. The current MSRV is 1.70, so we can now use `HashSet::from` and get rid of this macro.

https://github.com/GothenburgBitFactory/taskchampion/blob/e725a38c991dd32fd5f3b9b2daa4644b7dbbc9d2/taskchampion/src/macros.rs#L3-L17

koleesch:
I'll take this ticket

koleesch:
It is currently open.

djmitche:
Oops!

PR #419, opened 2024-07-08
replace set! with hashset::from

koleesch:
I hope, I do the right. This is my very first contribution to an opensource project.

The issue is the following https://github.com/GothenburgBitFactory/taskchampion/issues/415


djmitche (approved):
The patch looks good! Typically the PR description should say what it does without the context of the relevant issue, and the PR body should link to the issue (as you've done here).

As a quick tip, if you write "Fixes #415" then GitHub will automatically close the issue when the PR is merged.

djmitche:
Thanks! That was a good choice to practice making a PR :)

PR #423, opened 2024-07-16
Define Operations as a sequence of Operation-s

djmitche:
This is the next step in #372. This `Operations` type will eventually be the only way to make changes to tasks, so it has some nice usability bits like IntoIterator.


djmitche:
If you want a look at what the next few PRs are: https://github.com/GothenburgBitFactory/taskchampion/compare/main...djmitche:issue372-operations?expand=1

ryneeverett (approved):

PR #424, opened 2024-07-17
Add `BasicTask` and link it to Replica.

djmitche:
This adds a new, lower-level task API, without changing the existing `Task` and `TaskMut` types.

For the moment, I want to stick to non-breaking changes, and that precludes using the name `Task` for this type. We can consider renaming this later (rust-analyzer makes it really easy!) before a release.

Another step in #372.

### DIFF
--- a/taskchampion/src/replica.rs
+++ b/taskchampion/src/replica.rs
@@ -6,6 +6,7 @@ use crate::storage::{Storage, TaskMap};
 use crate::task::{Status, Task};
 use crate::taskdb::TaskDb;
 use crate::workingset::WorkingSet;
+use crate::BasicTask;
 use anyhow::Context;
 use chrono::{Duration, Utc};
 use log::trace;
@@ -184,7 +185,18 @@ impl Replica {
             .map(move |tm| Task::new(uuid, tm, depmap)))
     }
 
+    /// get an existing task by its UUID, as a [`BasicTask`](crate::BasicTask).
+    pub fn get_basic_task(&mut self, uuid: Uuid) -> Result<Option<BasicTask>> {
+        Ok(self
+            .taskdb
+            .get_task(uuid)?
+            .map(move |tm| BasicTask::new(uuid, tm)))
+    }
+
     /// Create a new task.
+    ///
+    /// This uses the high-level task interface. To create a task with the low-level
+    /// interface, use [`BasicTask::create`](crate::BasicTask::create).
     pub fn new_task(&mut self, status: Status, description: String) -> Result<Task> {
         let uuid = Uuid::new_v4();
         self.add_undo_point(false)?;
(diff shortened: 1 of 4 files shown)


taskchampion/src/task/basictask.rs
```diff
@@ -0,0 +1,254 @@
...
+/// This type presents a low-level interface consisting only of a key/value map. Interpretation of
+/// fields is up to the user, and modifications both modify the [`BasicTask`] and create one or
+/// more [`Operation`](crate::Operation) values that can later be committed to the replica.
+///
+/// This interface is intended for sophisticated applications like Taskwarrior which give meaning
+/// to key and values themselves. Use [`Task`](crate::Task) for a higher-level interface with
+/// methods to update status, set tags, and so on.
+///
+/// Wrappers of TaskChampion in other languages should only expose this interface, as the
+/// higher-level interface can be defined in terms of this one.
```
ryneeverett: The conclusion doesn't quite follow from the premise here. You explain why the high-level interface isn't *necessary* but that doesn't explain why accessing it from another language is *problematic*.

taskchampion/src/task/basictask.rs
```diff
@@ -0,0 +1,254 @@
...
+#[derive(Debug, PartialEq, Eq, Clone)]
+pub struct BasicTask {
+    uuid: Uuid,
+    pub(crate) taskmap: TaskMap,
```
ryneeverett: What's the rationale for making `taskmap` crate-public here? I assume it's because the high-level interface needs it but it isn't obvious to me why the high-level interface can't make do with the impl methods.

taskchampion/src/task/basictask.rs
```diff
@@ -0,0 +1,254 @@
...
+        });
+    }
+
+    /// Delete this task.
+    ///
+    /// Note that this is different from setting status to [`Deleted`](crate::Status::Deleted):
+    /// the resulting operation removes the task from the database.
+    ///
+    /// Deletion may interact poorly with modifications to the same task on other replicas. For
+    /// example, if a task is deleted on replica 1 and its description modified on replica 1, then
```
ryneeverett:
```suggestion
    /// example, if a task is deleted on replica 1 and its description modified on replica 2, then
```

taskchampion/src/replica.rs
```diff
@@ -201,6 +213,7 @@ impl Replica {
     /// Create a new, empty task with the given UUID.  This is useful for importing tasks, but
     /// otherwise should be avoided in favor of `new_task`.  If the task already exists, this
     /// does nothing and returns the existing task.
+    #[deprecated(since = "0.7.0", note = "please use BasicTask instead")]
```
ryneeverett, line 216: The deprecation seems a little premature since you can't import a `BasicTask` into a `Replica` yet because there isn't a mechanism to commit `Operations` into a `Replica` yet, right? (Feel free to let this slide if that's in a subsequent PR I haven't reviewed yet.)

ryneeverett (approved):

taskchampion/src/replica.rs
ryneeverett: Never mind, I see that's literally the next PR.

taskchampion/src/task/basictask.rs
djmitche: I'll just remove this sentence. My thinking was in terms of Task/TaskMut, where the idea that TaskMut contains an `&mut Replica`, and thus precludes any access to the replica while it exists, is difficult to handle correctly in other languages. But if we get rid of that interface, then it's not a problem.

taskchampion/src/task/basictask.rs
djmitche: Indeed, this isn't necessary. I suspect I added this when getting tests to run and didn't remember to remove it..

PR #425, opened 2024-07-17
Support committing `Operations`

djmitche:
A full set of Operations can be committed in one transaction, complete with updates to the task properties and the working set.

The next in #372. This has a merge conflict with #424 in the `use` declarations in `replica.rs` but it is trivially resolved.

### DIFF
--- a/taskchampion/src/taskdb/apply.rs
+++ b/taskchampion/src/taskdb/apply.rs
@@ -2,6 +2,44 @@ use crate::errors::{Error, Result};
 use crate::operation::Operation;
 use crate::server::SyncOp;
 use crate::storage::{StorageTxn, TaskMap};
+use crate::Operations;
+
+/// Apply `operations` to the dataabase, in the given single transaction.
+///
+/// This updates the set of tasks in the database, but does not modify the list of operations.
+///
+/// The transaction is not committed.
+pub(super) fn apply_operations(txn: &mut dyn StorageTxn, operations: &Operations) -> Result<()> {
+    for operation in operations {
+        match operation {
+            Operation::Create { uuid } => {
+                txn.create_task(*uuid)?;
+            }
+            Operation::Delete { uuid, .. } => {
+                txn.delete_task(*uuid)?;
+            }
+            Operation::Update {
+                uuid,
+                property,
+                value,
+                ..
+            } => {
+                // TODO: could cache this value from op to op
+                let task = txn.get_task(*uuid)?;
+                if let Some(mut task) = task {
+                    if let Some(v) = value {
+                        task.insert(property.clone(), v.clone());
+                    } else {
+                        task.remove(property);
+                    }
+                    txn.set_task(*uuid, task)?;
+                }
+            }
+            Operation::UndoPoint => {}
+        }
+    }
+    Ok(())
+}
 
 /// Apply the given SyncOp to the replica, updating both the task data and adding a
 /// Operation to the list of operations.  Returns the TaskMap of the task after the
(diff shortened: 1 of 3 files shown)


taskchampion/src/taskdb/mod.rs
```diff
@@ -31,6 +34,54 @@ impl TaskDb {
         TaskDb::new(Box::new(InMemoryStorage::new()))
     }
 
+    /// Apply `operations` to the dataabase, in a single transaction.
```
ryneeverett:
```suggestion
    /// Apply `operations` to the database in a single transaction.
```

taskchampion/src/taskdb/mod.rs
```diff
@@ -31,6 +34,54 @@ impl TaskDb {
...
+    ///
+    /// The resulting operations will be appended to the list of local operations, and the
```
ryneeverett:
```suggestion
    /// The operations will be appended to the list of local operations, and the
```

It's all the operations passed into the function, right?

taskchampion/src/taskdb/mod.rs
```diff
@@ -31,6 +34,54 @@ impl TaskDb {
...
+        for operation in &operations {
+            if add_to_working_set(operation) {
+                if let Some(uuid) = match operation {
+                    Operation::Create { uuid } => Some(*uuid),
+                    Operation::Update { uuid, .. } => Some(*uuid),
+                    Operation::Delete { uuid, .. } => Some(*uuid),
+                    _ => None,
+                } {
+                    to_add.push(uuid);
+                }
```
ryneeverett: I suspect you did not mean to use `if let` *and* `match`. i'm surprised this is valid rust.

taskchampion/src/taskdb/mod.rs
```diff
@@ -31,6 +34,54 @@ impl TaskDb {
...
+            }
+        }
+        let mut in_working_set: HashSet<Uuid> =
+            txn.get_working_set()?.iter().filter_map(|u| *u).collect();
```
ryneeverett: Seems like `working_set` would be a more appropriate name for this variable.

taskchampion/src/taskdb/mod.rs
```diff
@@ -31,6 +34,54 @@ impl TaskDb {
...
+        for uuid in to_add {
+            if !in_working_set.contains(&uuid) {
```
ryneeverett: The `add_to_working_set` passed by `Replica.commit_operations` is only going to evaluate to true if the task was not already pending or recurring and if it was not already pending it should not be in the working set, right? I'm not exactly sure how the `recurring` type works with respect to the working set so maybe this check is just for recurrence? On it's surface it seems like this if-statement might always evaluate true so could you comment on why it is needed?

taskchampion/src/taskdb/mod.rs
```diff
@@ -191,6 +242,87 @@ mod tests {
...
+                Operation::Create { uuid: uuid1 },
+                Operation::Create { uuid: uuid2 },
+                Operation::Create { uuid: uuid3 },
+                Operation::Create { uuid: uuid2 },
+                Operation::Create { uuid: uuid3 },
+            ]
+        );
+
+        // uuid2 was added to the working set, once, and uuid3 was not.
+        assert_eq!(db.working_set()?, vec![None, Some(uuid1), Some(uuid2)],);
```
ryneeverett, line 321: Does the working set always start with a `None`?

taskchampion/src/taskdb/apply.rs
```diff
@@ -2,6 +2,44 @@ use crate::errors::{Error, Result};
 use crate::operation::Operation;
 use crate::server::SyncOp;
 use crate::storage::{StorageTxn, TaskMap};
+use crate::Operations;
+
+/// Apply `operations` to the dataabase, in the given single transaction.
```
ryneeverett:
```suggestion
/// Apply `operations` to the database in the given single transaction.
```

taskchampion/src/taskdb/apply.rs
```diff
@@ -2,6 +2,44 @@ use crate::errors::{Error, Result};
...
+            }
+            Operation::Update {
+                uuid,
+                property,
+                value,
+                ..
+            } => {
+                // TODO: could cache this value from op to op
+                let task = txn.get_task(*uuid)?;
+                if let Some(mut task) = task {
```
ryneeverett, line 33: So an Update operation is ignored if the uuid isn't already in the database, right? Shouldn't that be an error or at least some kind of warning?

(edit: I see apply_operations_update_does_not_exist tests for this behavior below.)

taskchampion/src/taskdb/mod.rs
djmitche: I'm not sure why it would be invalid? The match expression has type `Option<Uuid>` and the `if let Some(uuid) = .. { }` matches against that type. But I think I can use alternation in a single match expression to say the same thing more clearly.

taskchampion/src/taskdb/mod.rs
djmitche: Yes, tasks are considered to start at index 1 (matching Taskwarrior behavior).

taskchampion/src/taskdb/mod.rs
djmitche: Yeah, I want to be sure we don't add a task to the working set twice, even if the working set and task statuses somehow get out of alignment.

taskchampion/src/taskdb/apply.rs
djmitche: I'll add a docstring.

djmitche:
Ugh, that rebase didn't work.

taskchampion/src/taskdb/mod.rs
ryneeverett: Thanks, I do find that more clear. After much squinting my eyes can now parse the original expression but I probably wouldn't be able to do it again next week.

ryneeverett (approved):

PR #428, opened 2024-07-21
 Refactor to use BasicTask and Operations

djmitche:
This refactors a bunch of Replica methods and the high-level Task/TaskMut to use BasicTask and commit operations to achieve their goals.

I summarize all of the deprecations together in a later commit, with the breaking changes.

This is one of the "boring" bits for #372 - the next is, too, and then we get to the interesting breaking changes.

### DIFF
--- a/taskchampion/src/replica.rs
+++ b/taskchampion/src/replica.rs
@@ -6,7 +6,7 @@ use crate::storage::{Storage, TaskMap};
 use crate::task::{Status, Task};
 use crate::taskdb::TaskDb;
 use crate::workingset::WorkingSet;
-use crate::{BasicTask, Error};
+use crate::{TaskData, Error};
 use anyhow::Context;
 use chrono::{Duration, Utc};
 use log::trace;
@@ -58,7 +58,7 @@ impl Replica {
     /// Update an existing task.  If the value is Some, the property is added or updated.  If the
     /// value is None, the property is deleted.  It is not an error to delete a nonexistent
     /// property.
-    #[deprecated(since = "0.7.0", note = "please use BasicTask instead")]
+    #[deprecated(since = "0.7.0", note = "please use TaskData instead")]
     pub fn update_task<S1, S2>(
         &mut self,
         uuid: Uuid,
@@ -72,7 +72,7 @@ impl Replica {
         let value = value.map(|v| v.into());
         let property = property.into();
         let mut ops = self.make_operations();
-        let Some(mut task) = self.get_basic_task(uuid)? else {
+        let Some(mut task) = self.get_task_data(uuid)? else {
             return Err(Error::Database(format!("Task {} does not exist", uuid)));
         };
         task.update(property, value, &mut ops);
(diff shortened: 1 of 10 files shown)


taskchampion/src/replica.rs
```diff
@@ -347,14 +363,23 @@ impl Replica {
     /// automatically when a change is made.  The `force` flag allows forcing a new UndoPoint
     /// even if one has already been created by this Replica, and may be useful when a Replica
     /// instance is held for a long time and used to apply more than one user-visible change.
+    #[deprecated(since = "0.7.0", note = "commit an Operation::UndoPoint instead")]
     pub fn add_undo_point(&mut self, force: bool) -> Result<()> {
         if force || !self.added_undo_point {
-            self.taskdb.add_undo_point()?;
```
ryneeverett, line 352: It appears that this is the only internal invocation of `taskdb.add_undo_point`. Do we want to keep that method?

taskchampion/src/replica.rs
```diff
@@ -347,14 +363,23 @@ impl Replica {
...
         }
         Ok(())
     }
 
+    fn add_undo_operation(&mut self, force: bool, operations: &mut Operations) {
+        if force || !self.added_undo_point {
+            operations.add(Operation::UndoPoint);
+            self.added_undo_point = true;
+        }
+    }
```
ryneeverett, line 381: It seems like every invocation of this method could be replaced with

```rs
        let mut ops = if self.added_undo_point { Operations::new() } else { Operations::new_with_undo_point() };
```

Unless you have further usages coming down the line...

taskchampion/src/task/task.rs
```diff
@@ -34,15 +35,14 @@ use uuid::Uuid;
 /// returns a TaskMut which can be used to modify the task.
 #[derive(Debug, Clone)]
 pub struct Task {
-    uuid: Uuid,
-    taskmap: TaskMap,
+    task: BasicTask,
```
ryneeverett: It seems like having some kind of naming condition for distinguishing between Task and BasicTask would be nice. The only idea coming to mind at the moment would be renaming BasicTask -> TaskData and calling this field `data`. :man_shrugging:

taskchampion/src/task/task.rs
```diff
@@ -165,17 +158,17 @@ impl Task {
...
     /// Determine whether this task is blocked -- that is, has at least one unresolved dependency.
     pub fn is_blocked(&self) -> bool {
-        self.depmap.dependencies(self.uuid).next().is_some()
+        self.depmap.dependencies(self.task.uuid()).next().is_some()
     }
 
     /// Determine whether this task is blocking -- that is, has at least one unresolved dependent.
     pub fn is_blocking(&self) -> bool {
-        self.depmap.dependents(self.uuid).next().is_some()
+        self.depmap.dependents(self.task.uuid()).next().is_some()
```
ryneeverett: Should we be using self.get_uuid() for consistency?

ryneeverett (commented):
- I'm seeing a few `self.task.task` in `TaskMut` and `task.task.task` in tests which I believe could be reduced by one ".task" due to Deref.
- I noticed that clippy is broken in CI again: ` Error: Unable to create clippy annotations! Reason: HttpError: Resource not accessible by integration` and there appear to be some relevant warnings. It sure would be nice if the check would fail.

taskchampion/src/replica.rs
djmitche: Yep, thanks!

`apply` gets removed in the next PR, in a more involved process, so I'll leave that for now.

taskchampion/src/replica.rs
djmitche: That doesn't update `added_undo_point`. But, I will refactor into a `self.make_operations()`.

taskchampion/src/task/task.rs
djmitche: Oh, I like `taskdata`!

taskchampion/src/task/task.rs
djmitche: Hm, I should probably just accept a slightly ugly name and change this `TaskData`/`BasicTask` method to `get_uuid` to match `Task::get_uuid`.

djmitche:
> I'm seeing a few self.task.task in TaskMut and task.task.task in tests which I believe could be reduced by one ".task" due to Deref.

This doesn't work because deref only works for method calls, so `self.task.foo()` uses the `task` property on `TaskMut`, of type `Task`, and then looks for a `foo` method on it.  Later PRs will address this, in particular when `Task` and `TaskMut` are unified into a single type.

> I noticed that clippy is broken in CI again

Oops, thanks for spotting that! A downside of a lot of rebasing.

I have a little more work to do here and will request review.

ryneeverett (approved):
Looks good. I don't entirely understand all the testing changes in the last commit but I presume they pertain to removing `apply` in the next PR.

djmitche:
> Looks good. I don't entirely understand all the testing changes in the last commit but I presume they pertain to removing `apply` in the next PR.

Yeah, a lot of the tests are in terms of SyncOp instead of Operation, and that gets addressed in the next PR ("Prefer Operation instead of SyncOp"). So, for this PR I've kept the SyncOp-related methods as test-only.

PR #429, opened 2024-07-22
Make clippy fail on warnings

djmitche:
As @ryneeverett noted on #428, as it stands clippy warnings aren't shown anywhere. This should "D"eny them. I've temporarily added an unused method to generate a warning.

### DIFF
--- a/taskchampion/src/replica.rs
+++ b/taskchampion/src/replica.rs
@@ -343,8 +343,6 @@ impl Replica {
         Ok(())
     }
 
-    fn unused() { todo!() }
-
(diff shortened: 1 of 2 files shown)


taskchampion/src/replica.rs
```diff
@@ -343,6 +343,8 @@ impl Replica {
         Ok(())
     }
 
+    fn unused() { todo!() }
+
```
djmitche:
```suggestion
```

djmitche (commented):
Yep, it worked.

ryneeverett (approved):
Thanks!

PR #431, opened 2024-07-25
Prefer `Operation` instead of `SyncOp` in tests

djmitche:
This is a test-only change: all modified code is `#[cfg(test)]`

Lots of tests used `SyncOp`, including some proptests. One of those is moved out to `tests/syncing-proptest.rs` and updated to use Replicas intead of TaskDb's. `op.rs` remains in terms of `SyncOp`, since it defines that type, but converts to `Operation` for tests.

With their uses in tests removed, two `apply_..` methods can finally be deleted.

This is the last "boring" change in #372. The next two will be breaking changes and have some open questions!

### DIFF
--- a/taskchampion/tests/syncing-proptest.rs
+++ b/taskchampion/tests/syncing-proptest.rs
@@ -1,0 +1,90 @@
+use pretty_assertions::assert_eq;
+use proptest::prelude::*;
+use taskchampion::{Operations, Replica, ServerConfig, StorageConfig, TaskData, Uuid};
+use tempfile::TempDir;
+
+#[derive(Debug, Clone)]
+enum Action {
+    Create,
+    Update(String, String),
+    Delete,
+    Sync,
+}
+
+fn action() -> impl Strategy<Value = Action> {
+    prop_oneof![
+        Just(Action::Create),
+        ("(description|project|due)", "(a|b|c)").prop_map(|(p, v)| Action::Update(p, v)),
+        Just(Action::Delete),
+        Just(Action::Sync),
+    ]
+}
+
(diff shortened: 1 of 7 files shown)


djmitche:
The one failing action here is #432 and not specific to this PR.

taskchampion/tests/syncing-proptest.rs
```diff
@@ -0,0 +1,90 @@
...
+        rep.sync(&mut server, false).unwrap()
+    }
+
+    let t0 = replicas[0].get_task_data(uuid).unwrap();
+    let t1 = replicas[1].get_task_data(uuid).unwrap();
+    let t2 = replicas[2].get_task_data(uuid).unwrap();
+    assert_eq!(t0, t1);
+    assert_eq!(t1, t2);
+}
+}
```
ryneeverett, line 90: Just out of curiosity, am I right to interpret the word "strategy" as "random choice" in the context of proptest? They use it pervasively in their docs and I'm not sure I know what they mean by it.

ryneeverett (approved):

taskchampion/tests/syncing-proptest.rs
djmitche: Yeah, but it's not quite random - proptest has a bunch of mechanisms to help it focus down on problematic test cases, so it's more like a transformation from a bit string to a test case, where that bit string it built in "smart" ways.

djmitche:
Yeah, but it's not quite random - proptest has a bunch of mechanisms to help it focus down on problematic test cases, so it's more like a transformation from a bit string to a test case, where that bit string it built in "smart" ways.

PR #435, opened 2024-07-26
[BREAKING] Unify Task and TaskMut

djmitche:
This breaking change removes the `TaskMut` type and moves its high-level setters to `Task`. `Task` derefs to `TaskData`, making the methods of that type available for `Task` values.

Several methods are deprecated, as detailed below. While these methods remain, they may be less performant than the recommended methods, either by reading unnecessary data from the database (e.g., to construct a dependency map) or by modifying the database in multiple transactions.

Breaking Changes:

 - `Task`:
   - All `TaskMut` setters are now methods on `Task`, and take `&mut ops` as their last argument.
   - `Task::into_mut` is removed.
   - `Task::delete` is deprecated. Use `Task::set_status` with `Status::Deleted` instead.

 - `Replica`:
   - `Replica::add_to_working_set` is removed - working set maintenance is now entirely automatic.
   - `Replica::new_task` is deprecated - prefer `Replica::create_task` and setting the `entry`, `description`, and `status` properties directly.
   - `Replica::import_task_with_uuid` is deprecated - prefer `Replica::create_task`.
   - `Replica::update_task` is deprecated - prefer `TaskData::update`.
   - `Replica::delete_task` is deprecated - prefer `TaskData::delete`.
   - Management of undo points in the replica, including `Replica::add_undo_point` and automatically adding undo points for various operations, is no longer supported; use `Operations::new_with_undo_point` to add one when necessary.

This is the first of two breaking changes for #372, which will necessitate a 0.7.0 release.

### DIFF
--- a/taskchampion/src/task/task.rs
+++ b/taskchampion/src/task/task.rs
@@ -2,7 +2,6 @@ use super::tag::{SyntheticTag, TagInner};
 use super::{utc_timestamp, Annotation, Status, Tag, Timestamp};
 use crate::depmap::DependencyMap;
 use crate::errors::{Error, Result};
-use crate::replica::Replica;
 use crate::storage::TaskMap;
 use crate::{Operations, TaskData};
 use chrono::prelude::*;
@@ -13,30 +12,27 @@ use std::rc::Rc;
 use std::str::FromStr;
 use uuid::Uuid;
 
-/* The Task and TaskMut classes wrap the underlying [`TaskMap`], which is a simple key/value map.
- * They provide semantic meaning to that TaskMap according to the TaskChampion data model.  For
- * example, [`get_status`](Task::get_status) and [`set_status`](TaskMut::set_status) translate from
- * strings in the TaskMap to [`Status`].
- *
- * The same approach applies for more complex data such as dependencies or annotations.  Users of
- * this API should only need the [`get_taskmap`](Task::get_taskmap) method for debugging purposes,
- * and should never need to make changes to the TaskMap directly.
- */
-
-/// A task, as publicly exposed by this crate.
+/// A task, with a high-level interface.
+///
+/// Building on [`crate::TaskData`], this type implements the task model, with ergonomic APIs to
+/// manipulate tasks without deep familiarity with the [task
+/// model](https://gothenburgbitfactory.org/taskchampion/tasks.html#keys). This type `Deref`s to
+/// [`crate::TaskData`] to support lower-level operations.
 ///
 /// Note that Task objects represent a snapshot of the task at a moment in time, and are not
 /// protected by the atomicity of the backend storage.  Concurrent modifications are safe,
 /// but a Task that is cached for more than a few seconds may cause the user to see stale
 /// data.  Fetch, use, and drop Tasks quickly.
-///
-/// This struct contains only getters for various values on the task. The
-/// [`into_mut`](Task::into_mut) method
-/// returns a TaskMut which can be used to modify the task.
 #[derive(Debug, Clone)]
 pub struct Task {
+    // The underlying task data.
     data: TaskData,
+
+    // The dependency map for this replica, for rapidly computing synthetic tags.
     depmap: Rc<DependencyMap>,
+
+    // True if an operation has alredy been emitted to update the `modified` property.
+    updated_modified: bool,
 }
 
 impl PartialEq for Task {
@@ -46,16 +42,18 @@ impl PartialEq for Task {
     }
 }
 
-/// A mutable task, with setter methods.
-///
-/// Most methods are simple setters and not further described.  Calling a setter will update the
-/// referenced Replica, as well as the included Task, immediately.
-///
-/// The [`Task`] methods are available on [`TaskMut`] via [`Deref`](std::ops::Deref).
-pub struct TaskMut<'r> {
-    task: Task,
-    replica: &'r mut Replica,
-    updated_modified: bool,
+impl std::ops::Deref for Task {
+    type Target = TaskData;
+
+    fn deref(&self) -> &Self::Target {
+        &self.data
+    }
+}
+
+impl std::ops::DerefMut for Task {
+    fn deref_mut(&mut self) -> &mut <Self as std::ops::Deref>::Target {
+        &mut self.data
+    }
 }
 
 /// An enum containing all of the key names defined in the data model, with the exception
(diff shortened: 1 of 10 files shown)


ryneeverett:
> `Task` derefs to `TaskData`, making the methods of that type available for `Task` values.

I haven't dug into it yet but I thought one of the premises of the refactor was to keep the high and low level interfaces distinct and that applications should be conscientious about which interface they're using. I'm not sure I still see this distinction as valuable with the current implementation, but without this distinction what value do we get from the extra wrapper layer of TaskData?

djmitche:
It's possible to ignore the `Task` type entirely and just use `TaskData`, with the advantage that the DependencyMap is never calculated. But, if you want the convenience methods like `set_status` or `add_annotation` or `is_blocking` (which depends on the DependencyMap) and are thus using the `Task` type, there are still cases where the `TaskData` methods might be useful, such as `task.properties` or `task.get_uuid`. The `Deref` implementation lets the user call these `TaskData` methods directly on the `Task` value.

That reminds me: `Task::delete` shadows `TaskData::delete`. The former marks the task as deleted, while the latter actually removes the task from the DB with a Delete operation. 
 * `pub fn Task::delete(&mut self, ops: &mut Operations) -> Result<(), Error>`
 * `pub fn TaskData::delete(self, operations: &mut Operations)`

I don't think this will cause confusion, and if it did it would be a shallow bug.

djmitche:
Hm, and as I'm working on using https://cxx.rs/ to bind this into Taskwarrior:
 * `delete` is a reserved word in C++ so it will need to be renamed, suggesting maybe renaming both methods might make sense
 * The move receiver (`self` instead of `&mut self`) isn't compatible with cxx, leading me to take the task by reference (`&self`) and clone it. So maybe that's too clever.

ryneeverett:
> It's possible to ignore the `Task` type entirely and just use `TaskData`, with the advantage that the DependencyMap is never calculated. But, if you want the convenience methods like `set_status` or `add_annotation` or `is_blocking` (which depends on the DependencyMap) and are thus using the `Task` type, there are still cases where the `TaskData` methods might be useful, such as `task.properties` or `task.get_uuid`. The `Deref` implementation lets the user call these `TaskData` methods directly on the `Task` value.

Yes, but as a Task user, don't you need to be conscientious when using TaskData methods since they don't commit operations? Doesn't the `Deref` make it less obvious that you're using the low-level interface and need to take responsibility for committing operations?

djmitche:
The new Task methods do not commit either: like Task data, they add operations to an `&mut Operations`.

But the Deref may not add a lot of value for users. Should I remove it?

djmitche:
I'll make a few changes here:
 - Remove the Deref/DerefMut implementations. While I don't think users would need to be conscientious, I also don't think it adds much value. These were widely used internally so this is a nontrivial change.
 - Deprecate `Task::delete`. It's just a simple wrapper around `task.set_status(Status::Deleted)` so users can simply do that.

ryneeverett:
> The new Task methods do not commit either: like Task data, they add operations to an `&mut Operations`.

I don't understand how this is the case. Pretty much all the Task methods call `set_value` which calls `self.replica.commit_operations(ops)?;` at the end.

Edit: I now see that this PR removes this line!

taskchampion/src/task/task.rs
```diff
@@ -508,71 +471,103 @@ impl<'r> TaskMut<'r> {
...
+        let mut task = replica.get_task(uuid).unwrap().unwrap();
+        assert(&task);
+        let mut ops = Operations::new();
+        modify(&mut task, &mut ops);
+        assert(&task);
+        println!("commiting operations from second call to modify function");
+        replica.commit_operations(ops).unwrap();
+        // check assertions on task loaded from storage again
+        let task = replica.get_task(uuid).unwrap().unwrap();
+        assert(&task);
```
ryneeverett: Is there a specific concern this second round of modify/assert/commit/assert is testing against or is it more a general robustness concern? Either way, separating this second round with a newline and an explanatory comment might be nice.

taskchampion/src/task/task.rs
```diff
@@ -833,233 +856,297 @@ mod test {
...
+            |task| {
+                let mut anns: Vec<_> = task.get_annotations().collect();
+                anns.sort();
+                assert_eq!(
+                    anns,
+                    vec![Annotation {
+                        entry: Utc.timestamp_opt(1635301883, 0).unwrap(),
+                        description: "left another message".into()
+                    }]
+                );
```
ryneeverett: Would it make more sense to assert that annotations are an empty Vec here?

taskchampion/src/task/task.rs
```diff
@@ -97,29 +80,27 @@ fn uda_tuple_to_string(namespace: impl AsRef<str>, key: impl AsRef<str>) -> Stri
 }
 
 impl Task {
-    pub(crate) fn new(uuid: Uuid, taskmap: TaskMap, depmap: Rc<DependencyMap>) -> Task {
+    pub(crate) fn new(data: TaskData, depmap: Rc<DependencyMap>) -> Task {
         Task {
-            data: TaskData::new(uuid, taskmap),
+            data,
```
ryneeverett, line 85: What is the rationale for this change?

taskchampion/src/task/task.rs
```diff
@@ -1115,117 +1206,156 @@ mod test {
...
+        with_mut_task(
+            |task, ops| {
+                task.add_dependency(dep1, ops).unwrap();
+                task.add_dependency(dep2, ops).unwrap();
+                task.remove_dependency(dep2, ops).unwrap();
+            },
+            |task| {
+                let deps = task.get_dependencies().collect::<Vec<_>>();
+                assert!(deps.contains(&dep1));
+            },
```
ryneeverett: This doesn't assert that dep2 was removed.

ryneeverett (approved):

djmitche:
> Edit: I now see that this PR removes this line!

You had me really worried & confused :)

taskchampion/src/task/task.rs
djmitche: Yeah, general robustness. We're "applying" the operation locally on the HashMap, so it makes sense to check that the same changes occur through all of the TaskDb machinery.

taskchampion/src/task/task.rs
djmitche: Yes! I had gotten the timestamps wrong in this test, so annotation deletion does indeed work.

taskchampion/src/task/task.rs
djmitche: IIRC,it was so I could call `Ok(Task::new(TaskData::create(uuid, ops), depmap))` without having to de-construct and re-construct that `TaskData`.

Issue #427, opened 2024-07-21
Make `commit_reversed_operations` more capable.


djmitche:
Right now, this method (known as `commit_undo_ops` before #372) just checks that the latest set of un-synchronized operations in the replica match the operations to undo. If that's the case, it removes them from the list of un-synchronized operations, reverses their effect on the tasks, and declares success. Otherwise, it fails.

This could be more robust -- if an operation does not precisely match the most recently recorded local operation, there's enough data to create a _new_, reversed operation (switch Create <--> Delete, swap `value` and `old_value` in Update) and commit that. Adding such support would be a big step on the way to allowing "undo" _after_ a sync has occurred.

I think we should fail the `commit_reversed_operations` if the reversed operation doesn't apply cleanly. So:
 - A reversed Create is a Delete, but if the task doesn't exist, that should fail.
 - A reversed Delete is a Create, but if the task still exists, that should fail.
 - A reversed Update specifies the `old_value` it expects to find, but if the value is different, it should fail.

PR #437, opened 2024-07-29
[BREAKING] Update undo support to use Operations

djmitche:
This updates the existing "undo" support to use `Operations`, while making some additional changes to look forward to more general undo support (#427 being the next step on that journey).

Breaking changes:

 - `Replica::get_undo_ops` is now `Replica::get_undo_operations` and returns an `Operations` value. It now returns the operations in the order they were applied, and includes the undo point operation, if one exists.
 - `Replica::commit_undo_ops` is now `Replica::commit_reversed_operations` and takes an `Operations` value as provided by `get_undo_operations`.

This is the second of the two planned breaking changes in #372.

### DIFF
--- a/taskchampion/src/operation.rs
+++ b/taskchampion/src/operation.rs
@@ -71,11 +71,6 @@ impl Operations {
         self.0.push(op);
     }
 
-    /// Determine if this set of operations is empty.
-    pub fn is_empty(&self) -> bool {
-        self.0.is_empty()
-    }
-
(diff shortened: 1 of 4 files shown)


ryneeverett (approved):
A notable change under the hood here is that UndoPoint's themselves are now reversed, right? That makes sense to me but just want to verify that was your intention.

djmitche:
Yes. The idea is to make commit_reversed_operations a little more generic by not assuming it is reversing to an undo point.

PR #438, opened 2024-07-30
Documentation clean-up

djmitche:
This fixes several warnings from `cargo doc` and other minor out-of-date or missing documentation.


ryneeverett (approved):
Should we have a github actions workflow to check for cargo doc warnings?

djmitche:
Yes, that's a good idea. I wrote this without wifi so I couldn't experiment with that!

PR #443, opened 2024-08-05
Use `Vec<Operation>` instead of newtype `Operations`

### LINKED ISSUE #372: Reconsider "level" of the Task/TaskMut interface

djmitche:
Exposing the full functionality of a Vec here is more useful than utility functions on a newtype, especially when building a Cxx bridge.

This is a big diff, but aside from the main change in `operation.rs`, consists almost entirely of changing `ops.add` to `ops.push`, with a few related test changes.

This will be the last change for #372, and without objection I will release 0.7.0 when this is merged.

### DIFF
--- a/taskchampion/src/replica.rs
+++ b/taskchampion/src/replica.rs
@@ -422,6 +422,7 @@ impl Replica {
         let mut ops = Operations::new();
         if !self.added_undo_point {
             ops.push(Operation::UndoPoint);
+            self.added_undo_point = true;
         }
         ops
     }
--- a/taskchampion/src/replica.rs
+++ b/taskchampion/src/replica.rs
@@ -405,11 +405,11 @@ impl Replica {
     /// instance is held for a long time and used to apply more than one user-visible change.
     #[deprecated(
         since = "0.7.0",
-        note = "Use `Operations::new_with_undo_point` instead."
+        note = "Push an `Operation::UndoPoint` onto your `Operations` instead."
     )]
     pub fn add_undo_point(&mut self, force: bool) -> Result<()> {
         if force || !self.added_undo_point {
-            let ops = Operations::new_with_undo_point();
+            let ops = vec![Operation::UndoPoint];
             self.commit_operations(ops)?;
             self.added_undo_point = true;
         }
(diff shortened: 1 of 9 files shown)


taskchampion/src/replica.rs
```diff
@@ -419,12 +419,11 @@ impl Replica {
     /// Make a new `Operations`, with an undo operation if one has not already been added by
     /// this `Replica` insance
     fn make_operations(&mut self) -> Operations {
-        if self.added_undo_point {
-            Operations::new()
-        } else {
-            self.added_undo_point = true;
```
ryneeverett, line 425: Isn't this line still needed? (`self.added_undo_point = true;`)

ryneeverett (approved):
You know, I almost suggested this when reviewing Operations the first time -- it seemed like an extra abstraction which wasn't "buying" us much over a Vec. But it wasn't a terribly expensive abstraction so I figured I'd let you have it -- but I guess cxx isnt' so generous. :)

taskchampion/src/replica.rs
djmitche: Oh, yes!!

PR #444, opened 2024-08-08
Expose sqlite bundling as a feature

jcgruenhage:
Linux distributions usually don't want the libraries to be statically linked. Before the repos for taskchampion and taskwarrior were split, it was easy to get rid of the bundled sqlite by patching, but with taskchampion moved to a separate repository this became more complicated.

I do get why you want to default to bundling sqlite, so I made it so that the default behaviour is still that it's being bundled.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,7 +22,7 @@ proptest = "^1.5.0"
 regex = "^1.10.2"
 ring = "0.17"
 rstest = "0.17"
-rusqlite = { version = "0.29", features = ["bundled"] }
+rusqlite = { version = "0.29"}
 serde_json = "^1.0"
 serde = { version = "^1.0.147", features = ["derive"] }
 strum = "0.25"
--- a/taskchampion/Cargo.toml
+++ b/taskchampion/Cargo.toml
@@ -12,8 +12,10 @@ edition = "2021"
 rust-version = "1.70.0"
 
 [features]
-default = ["server-sync", "server-gcp"]
+default = ["sync", "bundled"]
 
+# Support for all sync solutions
+sync = ["server-sync", "server-gcp"]
 # Support for sync to a server
 server-sync = ["encryption", "dep:ureq", "dep:url"]
 # Support for sync to GCP
@@ -22,6 +24,8 @@ server-gcp = ["cloud", "encryption", "dep:google-cloud-storage", "dep:tokio"]
 encryption = ["dep:ring"]
 # (private) Generic support for cloud sync
 cloud = []
+# static bundling of dependencies
+bundled = ["rusqlite/bundled"]
 
 [package.metadata.docs.rs]
 all-features = true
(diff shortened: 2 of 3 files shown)


djmitche (changes requested):
I'm happy to merge this, although it needs a bit of documentation of the feature flags in the [module docstring](https://docs.rs/taskchampion/latest/taskchampion/#feature-flags).

I suppose with this change, the patch to the taskwarrior build becomes easier, by just adding this flag?

And while I'm asking, is there a specific distro that you're working on?

djmitche (approved):
Thanks!!

jcgruenhage:
I'll start with the distro, this is Chimera Linux, which has been shipping taskwarrior since a week after the 3.0.0 release. Taskwarrior 3.0.0 was the first release shipped by Chimera Linux, earlier versions were not packaged.

Yes, this change would make it easier to build taskwarrior again. When taskchampion and taskwarrior lived in the same repo, it was reasonably easy to do, but with different repositories this became a bit harder. Previously, it would look like this: https://github.com/chimera-linux/cports/blob/master/user/taskwarrior/patches/0003-unvendor-sqlite.patch

Chimera Linux builds packages in phases (see https://github.com/chimera-linux/cports/blob/master/Packaging.md#phases for details), and the interesting ones for why this patch is necessary are `prepare` and `patch`. Usually when building rust tools, `prepare` runs `cargo vendor` to fetch the sources of all dependencies, and `patch` can then also patch the dependencies if needed. This does *not* work if the patches also change the `Cargo.toml`, but this is something that can be worked around. Now, if the `Cargo.toml` in a dependency changes, this becomes a whole lot more complicated.

What I've done in the past (other distro, other tool, but same problem) is forking the library and replacing the dependency. An example for this is https://github.com/jcgruenhage/void-packages/blob/58a0618f0529dc77dc3de6771d44876d94dd6dbc/srcpkgs/nushell/patches/0001-use-system-libraries.patch#L60-L61. That is super tedious though, and getting feature flags to toggle bundling at the binary level (so in this case taskwarrior) would be best, because then patching can be completely avoided, at least for bundling.

djmitche:
Interesting -- thanks for the details, and thanks for packaging Taskwarrior :)

jcgruenhage:
Always! Thanks for your support and thanks for your work on taskchampion/taskwarrior!

I might need help down the line to get to that final goal of not needing patches for taskwarrior anymore. There are things that we're patching that I can upstream easily, but the `bundled` patch needs cmake changes that are just slightly beyond my abilities right now. I don't know much about cmake yet, taskwarrior is the first project where I'm diving a little deeper into the build system that uses cmake.

But all in due time: the taskchampion changes will need some time to land in a release anyway, so I'll first work on the changes that are easier to upstream and don't have dependencies like that.

djmitche:
Sounds good. I'm not much good at CMake either, but we can all puzzle it out together.

PR #446, opened 2024-08-19
docs(tasks.md): fix broken emphasis

gagath:
In commit 2ab47269afebb8a541c71b180bfb5d16857b29a8 the emphasis for this word was broken. It currently appears as a single underscore on the published book website.

image


djmitche (approved):
Nice catch, thank you!

Issue #373, opened 2022-08-28
Expose operations from TaskChampion


djmitche:
## Background

Every change to a task is represented internally as an "Operation"
https://github.com/GothenburgBitFactory/taskwarrior/blob/41992d484909bd865bc252e99e588c5c3f37c71a/taskchampion/taskchampion/src/storage/op.rs#L10-L38
This is how we accomplish `task sync` -- different replicas share the set of operations they have applied, and resolve any conflicts in a reasonable, consistent way.

Note that a single command-line invocation may produce several operations. For example
```
task 123 start +inprogress prio:M
```
would create four `Update` operations: one to set the `start` property to the current time, one to add the `tag_inprogress` property, one to set the `priority` property to `M` and one to update the `modified` property to the current time.

There are also `UndoPoint` operations, which do not actually change anything but act as a signpost for how many operations to revert when running `task undo` -- that command basically reverts operations until it reaches an `UndoPoint`.  The `UndoPoint` operations are not sent to the server.

Once operations are successfully sent to the server, they are no longer stored locally. This avoids the local task database growing forever.

Currently, none of this is visible in Taskchampion's public API. This bug would change that.

## Motivation

We'll need this for a few reasons:

The `task undo`  command _previously_ showed a "diff" of what it would change, and asked the user for permission. This is not possible with the current Taskchampion API (see https://github.com/GothenburgBitFactory/taskwarrior/commit/4b814bc602a365342a7ebb101c8d156f8cd4a15a). But if there was a way to query Taskchampion to say "hey, what are the operations back to the latest `UndoPoint`?" then it could generate a diff view from that. 

The `task info` command has historically used the undo information to show the history for a particular task. This operated by scanning the entire undo history for records matching that task's uuid, and formatting those nicely for display. Again, that's currently impossible (support was removed in GothenburgBitFactory/taskwarrior#3060). If there was a way to query Taskchampion to say "hey, what operations have happened for this task?" then we could bring back that functionality.

[...]

PR #449, opened 2024-09-02
Add an 'operations.uuid' column, generated and virtual

djmitche:
This column is not stored separately, but allows indexing operations by uuid without further modification or redundant storage. DB's have some cool features!

This is part of ongoing work toward #373. You can see the whole branch in https://github.com/GothenburgBitFactory/taskchampion/compare/main...djmitche:issue373.

### DIFF
--- a/taskchampion/src/operation.rs
+++ b/taskchampion/src/operation.rs
@@ -43,6 +43,16 @@ impl Operation {
     pub fn is_undo_point(&self) -> bool {
         self == &Self::UndoPoint
     }
+
+    /// Get the UUID for this function, if it has one.
+    pub fn get_uuid(&self) -> Option<Uuid> {
+        match self {
+            Operation::Create { uuid: u } => Some(*u),
+            Operation::Delete { uuid: u, .. } => Some(*u),
+            Operation::Update { uuid: u, .. } => Some(*u),
+            Operation::UndoPoint => None,
+        }
+    }
 }
 
 /// Operations are a sequence of [`Operation`] values, which can be committed in a single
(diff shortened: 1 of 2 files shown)


ryneeverett:
I'm not sure how exactly this fits in the latest strategy you've described in #373 but it does suggest the possibility of passing references to "losing" operations rather than values. The duplication of operations -- which basically have an unbounded size -- worries me.

ryneeverett (approved):

djmitche:
I'm not sure what you mean by "duplication of operations". The discussion in #373 does talk about values being duplicated, in order to indicate which operation "won" while also including "losing" operations. But that's not particularly relevant to this PR, which just allows indexing operations by task uuid.

I'm also confused by references vs. values -- that makes sense in Rust terms, but none of that is happening in this PR. And I think SQL is entirely pass-by-value, although I've not thought about that too hard.

If you meant "referencing" an operation by some unique ID, I'm not sure how we would establish that unique ID across replicas, and also ensure that the referent was always present on a replica where the reference was present.

ryneeverett:
> If you meant "referencing" an operation by some unique ID, I'm not sure how we would establish that unique ID across replicas, and also ensure that the referent was always present on a replica where the reference was present.

Yeah, that is what I meant. Sorry I was using rust terminology analogously, that was cruel.

I already had in mind that a modification of the sync protocol employing operation uuid's might escape some of the obstacles we were discussing, including duplication and the invariant that the replica state is the result of all operations by separating the data of operations from the concept of "all operations". "All operations", with respect to the invariant, could be an array of uuids. And with global uuid's, duplication could probably be eliminated and transmission could be limited to the uuid rather than the whole operation. I'm just spitballing a a significant change here though and hoping you have a more elegant solution.

ryneeverett:
Ok, so after reviewing #450 I now see that I had missed that these uuid's don't belong to the operation itself but the task it operates on so my suggestion was irrelevant to this PR.

djmitche:
Ah, I see!

PR #451, opened 2024-09-03
Replace `Storage::set_operations` with `sync_complete` and `remove_operation`.`

djmitche:
When #373 is complete, we will be keeping the set of operations around, so clearing it or setting it to a specific set of operations no longer makes sense. Instead, we push or pop operations, and track when the current set of operations has been synced.

### DIFF
--- a/taskchampion/src/taskdb/undo.rs
+++ b/taskchampion/src/taskdb/undo.rs
@@ -68,26 +68,32 @@ fn reverse_ops(op: Operation) -> Vec<SyncOp> {
 /// have not yet been synchronized, and will return `false` if this is not the case.
 pub fn commit_reversed_operations(txn: &mut dyn StorageTxn, undo_ops: Operations) -> Result<bool> {
     let mut applied = false;
-    let mut local_ops = txn.operations().unwrap();
+    let local_ops = txn.operations().unwrap();
     let mut undo_ops = undo_ops.to_vec();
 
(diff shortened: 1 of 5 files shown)


taskchampion/src/taskdb/undo.rs
```diff
@@ -90,16 +90,16 @@ pub fn commit_reversed_operations(txn: &mut dyn StorageTxn, undo_ops: Operations
```
ryneeverett, line 95: We could do a bit of cleanup in this method now that `local_ops` is not being used to replace the operations:

- Remove `local_ops.truncate(new_len);`.
- I believe we could delete the entire section above except for `undo_ops.reverse();` if we optimistically proceed with reversal and log those messages. Something like:

```rs
    let Ok(ok) = txn.remove_operation(op) else {
        info!("Undo failed: concurrent changes to the database occurred.");
        debug!(
            "local_undo_ops={:#?}\nundo_ops={:#?}",
            local_undo_ops, undo_ops
        );
        return Ok(applied);
    };
```

ryneeverett (approved):

taskchampion/src/taskdb/undo.rs
djmitche: The `remove_operation` might fail for other reasons than inconsistent operations, so that could lead to incorrect error messages. I think it's best to still compare the lists before beginning. But, you're right that it can be simplified.

Issue #452, opened 2024-09-07
Add `Replica::pending_tasks` and `pending_task_data`


djmitche:
We already have [`Replica::all_tasks`](https://docs.rs/taskchampion/latest/taskchampion/struct.Replica.html#method.all_tasks) and [`all_task_data`](https://docs.rs/taskchampion/latest/taskchampion/struct.Replica.html#method.all_task_data), so it may make sense to add `Replica::pending_tasks` and `pending_task_data` which could then join on the working_set table (as opposed to taking an arbitrary list of UUIDs).

I suspect the implementation of this would be relatively straightforward, and based on existing examples at both the `Replica` and underlying `TaskDB` levels in TaskChampion. The connection to C++ is now much simpler than it used to be, [using cxx](https://github.com/GothenburgBitFactory/taskwarrior/blob/72f9cd91a519034b592fc3069da1e764776a203b/src/taskchampion-cpp/src/lib.rs), so that should also be relatively straightforward. And then calling the result from TDB2 is also pretty easy. The tricky bit is implementing it in TaskChampion and then developing against that in Taskwarrior, since they are different repos -- but I can help with that. This is definitely worth doing even if it's not the biggest contributor to poor performance. Do you want to give it a shot? If so, let's open a new issue to track the work.

_Originally posted by @djmitche in https://github.com/GothenburgBitFactory/taskwarrior/issues/3329#issuecomment-2335007219_

Scandiravian:
Hey @djmitche,

I've started looking into this and have written a simple implementation of `pending_tasks` with the signature `pub fn pending_tasks(&mut self) -> Result<Vec<Task>> `. This was to match this part of the cpp function: https://github.com/GothenburgBitFactory/taskwarrior/blob/40ea3f2f5418ab5795e2aa0e73a1206c1d18cc1e/src/TDB2.cpp#L355-L367

I'm not entirely sure this is the way you'd prefer it was done, so I'd like to open a draft so you can have a look before I start on the `pending_task_data` function if that's alright with you?

djmitche:
Sure, that sounds great!

metent:
Hi @djmitche, I am planning to incorporate taskchampion for my tasks app. I was reading through taskchampion's docs and source code when I noticed the lack of a function which returns tasks based on an applied filter. I skimmed through taskwarrior's source to see how they implemented it and I believe they iterate through all tasks and check if each task satisfy the criteria rather than running an sqlite query.

[...]

PR #454, opened 2024-09-11
implement Replica::pending_tasks

Scandiravian:
This is a first attempt at #452. 

I've implemented two different approaches in the storage layer and I'm looking for feedback on whether either of these, or a third approach, would be most optimal. 

Currently the `taskdb` layer use the more "naive" approach, which follows the original c++ code more closely. 

The alternative approach would use the `get_tasks_by_status` to query for `pending` tasks. I am however not sure whether there could potentially be an inconsistency between the `working_set` and the sqlite database, which would make that solution unfeasible.

### DIFF
--- a/taskchampion/src/replica.rs
+++ b/taskchampion/src/replica.rs
@@ -106,6 +106,35 @@ impl Replica {
         self.taskdb.all_task_uuids()
     }
 
+    /// Get an array containing all pending tasks
+    pub fn pending_tasks(&mut self) -> Result<Vec<Task>> {
+        let depmap = self.dependency_map(false)?;
+        let res = self
+            .pending_task_data()?
+            .into_iter()
+            .map(|taskdata| Task::new(taskdata, depmap.clone()))
+            .collect();
+
+        Ok(res)
+    }
+
+    pub fn pending_task_data(&mut self) -> Result<Vec<TaskData>> {
+        let uuids = self
+            .working_set()?
+            .iter()
+            .map(|(_, uuid)| uuid)
+            .collect::<Vec<Uuid>>();
+
+        let res = self
+            .taskdb
+            .get_tasks(uuids)?
+            .into_iter()
+            .map(|(uuid, taskmap)| TaskData::new(uuid, taskmap))
+            .collect::<Vec<_>>();
+
+        Ok(res)
+    }
+
     /// Get the "working set" for this replica.  This is a snapshot of the current state,
     /// and it is up to the caller to decide how long to store this value.
     pub fn working_set(&mut self) -> Result<WorkingSet> {
@@ -636,6 +665,30 @@ mod tests {
         assert_eq!(all_tasks.len(), 2);
         assert_eq!(all_tasks.get(&uuid1).unwrap().get_uuid(), uuid1);
         assert_eq!(all_tasks.get(&uuid2).unwrap().get_uuid(), uuid2);
+    }
+
+    #[test]
+    fn pending_tasks() {
+        let mut rep = Replica::new_inmemory();
+
+        let (uuid1, uuid2, uuid3) = (Uuid::new_v4(), Uuid::new_v4(), Uuid::new_v4());
+        let mut ops = Operations::new();
+
+        let mut t1 = rep.create_task(uuid1, &mut ops).unwrap();
+        t1.set_status(Status::Pending, &mut ops).unwrap();
+
+        let mut t2 = rep.create_task(uuid2, &mut ops).unwrap();
+        t2.set_status(Status::Pending, &mut ops).unwrap();
+
+        let mut t3 = rep.create_task(uuid3, &mut ops).unwrap();
+        t3.set_status(Status::Completed, &mut ops).unwrap();
+
+        rep.commit_operations(ops).unwrap();
+
+        let all_tasks = rep.pending_tasks().unwrap();
+        assert_eq!(all_tasks.len(), 2);
+        assert_eq!(all_tasks.get(0).unwrap().get_uuid(), uuid1);
+        assert_eq!(all_tasks.get(1).unwrap().get_uuid(), uuid2);
     }
 
     #[test]
--- a/taskchampion/src/replica.rs
+++ b/taskchampion/src/replica.rs
@@ -119,15 +119,9 @@ impl Replica {
     }
 
     pub fn pending_task_data(&mut self) -> Result<Vec<TaskData>> {
-        let uuids = self
-            .working_set()?
-            .iter()
-            .map(|(_, uuid)| uuid)
-            .collect::<Vec<Uuid>>();
-
         let res = self
             .taskdb
-            .get_tasks(uuids)?
+            .get_pending_tasks()?
             .into_iter()
             .map(|(uuid, taskmap)| TaskData::new(uuid, taskmap))
             .collect::<Vec<_>>();
--- a/taskchampion/src/replica.rs
+++ b/taskchampion/src/replica.rs
@@ -679,10 +679,10 @@ mod tests {
 
         rep.commit_operations(ops).unwrap();
 
-        let all_tasks = rep.pending_tasks().unwrap();
-        assert_eq!(all_tasks.len(), 2);
-        assert_eq!(all_tasks.get(0).unwrap().get_uuid(), uuid1);
-        assert_eq!(all_tasks.get(1).unwrap().get_uuid(), uuid2);
+        let pending_tasks = rep.pending_tasks().unwrap();
+        assert_eq!(pending_tasks.len(), 2);
+        assert_eq!(pending_tasks.get(0).unwrap().get_uuid(), uuid1);
+        assert_eq!(pending_tasks.get(1).unwrap().get_uuid(), uuid2);
     }
 
     #[test]
--- a/taskchampion/src/storage/inmemory.rs
+++ b/taskchampion/src/storage/inmemory.rs
@@ -54,6 +54,10 @@ impl<'t> StorageTxn for Txn<'t> {
             .get_working_set()?
             .iter()
             .filter_map(|uuid| {
+                // Since uuid is wrapped in an Option and get(&inner_uuid)
+                // also returns an Option, the resulting type will be
+                // Option<Option<(Uuid, TaskMap)>>. To turn that into
+                // an Option<(Uuid, TaskMap)>, flatten is called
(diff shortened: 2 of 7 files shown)


djmitche:
Nice work! I think it would be unusual to want to get all completed or deleted tasks (and those cases would be covered by `all_tasks()`).

And, the set of all pending tasks and the working set can be slightly different in ways that might be significant. I think the best approach here would be joining with the working set to get the working-set tasks.

Scandiravian:
> Nice work! I think it would be unusual to want to get all completed or deleted tasks (and those cases would be covered by `all_tasks()`).
> 
> And, the set of all pending tasks and the working set can be slightly different in ways that might be significant. I think the best approach here would be joining with the working set to get the working-set tasks.

Thanks for the feedback. I've updated the PR so it hopefully reflect your suggestions. I moved some code from `pending_tasks` to `pending_tasks_data` when implementing the latter, but the functionality should be the same.

I have also fixed the CI failures, so I think this is ready for review. Let me know if you'd like any other changes to the code

djmitche:
> > Nice work! I think it would be unusual to want to get all completed or deleted tasks (and those cases would be covered by `all_tasks()`).
> > And, the set of all pending tasks and the working set can be slightly different in ways that might be significant. I think the best approach here would be joining with the working set to get the working-set tasks.
> 
> Thanks for the feedback. I've updated the PR so it hopefully reflect your suggestions. I moved some code from `pending_tasks` to `pending_tasks_data` when implementing the latter, but the functionality should be the same.
> 
> I have also fixed the CI failures, so I think this is ready for review. Let me know if you'd like any other changes to the code

This implementation looks solid and well-tested!

However, this still requires two separate queries to get the pending tasks: first to get the list of UUIDs in the pending set, and second to fetch each of those tasks. This involves sending that list of UUIDs from the DB to Rust and back. Could a single `get_pending_tasks` method do this in one query, with something like `SELECT tasks.* from tasks join working_set on..`?

Scandiravian:
That's really good feedback. I hadn't realised that the working set already exists in the database, thanks :smile: 

I'm guessing that there's somewhere else in the code that ensures only tasks with the "pending" status are added to the working_set?

I've added a new commit that changes the implementation to using a join to do it in a single query. Let me know if you would like any other changes. I'm personally not entirely happy with the `inmemory` implementation of `get_pending_tasks` as it's a bit too nested/complicated making it difficult to read

If you're happy with how the implementation is looking I'm going to squash the changes into a single commit before merging and give the code another pass to see if I can improve the readability+improve the doc strings

djmitche:
> I'm guessing that there's somewhere else in the code that ensures only tasks with the "pending" status are added to the working_set?

Yes, modifications to the `status` property trigger adjustments of the working set.
 
> I've added a new commit that changes the implementation to using a join to do it in a single query. Let me know if you would like any other changes. I'm personally not entirely happy with the `inmemory` implementation of `get_pending_tasks` as it's a bit too nested/complicated making it difficult to read
> 
> If you're happy with how the implementation is looking I'm going to squash the changes into a single commit before merging and give the code another pass to see if I can improve the readability+improve the doc strings

I'll have a look shortly!

taskchampion/src/replica.rs
```diff
@@ -638,6 +661,30 @@ mod tests {
...
+
+        let mut t2 = rep.create_task(uuid2, &mut ops).unwrap();
+        t2.set_status(Status::Pending, &mut ops).unwrap();
+
+        let mut t3 = rep.create_task(uuid3, &mut ops).unwrap();
+        t3.set_status(Status::Completed, &mut ops).unwrap();
+
+        rep.commit_operations(ops).unwrap();
+
+        let all_tasks = rep.pending_tasks().unwrap();
```
djmitche: Just for accuracy's sake, please call tihs `pending_tasks`..

taskchampion/src/storage/inmemory.rs
```diff
@@ -49,6 +49,24 @@ impl<'t> StorageTxn for Txn<'t> {
...
+            .filter_map(|uuid| {
+                uuid.map(|inner_uuid| {
+                    self.data_ref()
+                        .tasks
+                        .get(&inner_uuid)
+                        .map(|taskmap| (inner_uuid, taskmap.clone()))
+                })
+                .flatten()
+            })
+            .collect::<Vec<_>>();
```
djmitche, line 69: This is nice! It took me a bit to realize that `.get(..)` also returned an option, resulting in `Option<Option<(uuid, TaskMap)>>` and requiring the `.flatten()`.

taskchampion/src/storage/test.rs
```diff
@@ -267,6 +272,54 @@ pub(super) fn delete_task_exists(mut storage: impl Storage) -> Result<()> {
...
+            (
+                uuid2,
+                taskmap_with(vec![("num".to_string(), "2".to_string())]),
+            ),
+        ];
+        exp.sort_by(|a, b| a.0.cmp(&b.0));
+        tasks.sort_by(|a, b| a.0.cmp(&b.0));
+
+        assert_eq!(tasks, exp);
+    }
```
djmitche, line 323: I modified this test as follows, and it fails for sqlite. I think this is because of the `LEFT JOIN` - a task which matches no rows in the working set is included in the result set. I think the query should just be `JOIN` (and I think I led you astray in an earlier comment -- sorry!)

```diff
diff --git taskchampion/src/storage/test.rs taskchampion/src/storage/test.rs
index 390bf2f6c..9fe503fd3 100644
--- taskchampion/src/storage/test.rs
+++ taskchampion/src/storage/test.rs
@@ -268,38 +268,41 @@ pub(super) fn delete_task_exists(mut storage: impl Storage) -> Result<()> {
     {
         let mut txn = storage.txn()?;
         assert!(txn.delete_task(uuid)?);
     }
     Ok(())
 }
 
 pub(super) fn get_pending_tasks(mut storage: impl Storage) -> Result<()> {
-    let uuid1 = Uuid::new_v4();
-    let uuid2 = Uuid::new_v4();
+    let uuid1 = Uuid::new_v4(); // in working set
+    let uuid2 = Uuid::new_v4(); // in working set
+    let uuid3 = Uuid::new_v4(); // in working set but does not exist
+    let uuid4 = Uuid::new_v4(); // not in working set but exists
 
     {
         let mut txn = storage.txn()?;
         txn.add_to_working_set(uuid1)?;
         txn.add_to_working_set(uuid2)?;
-        txn.commit()?;
-    }
-
-    {
-        let mut txn = storage.txn()?;
+        txn.add_to_working_set(uuid3)?;
         assert!(txn.create_task(uuid1)?);
         txn.set_task(
             uuid1,
             taskmap_with(vec![("num".to_string(), "1".to_string())]),
         )?;
         assert!(txn.create_task(uuid2)?);
         txn.set_task(
             uuid2,
             taskmap_with(vec![("num".to_string(), "2".to_string())]),
         )?;
+        assert!(txn.create_task(uuid4)?);
+        txn.set_task(
+            uuid4,
+            taskmap_with(vec![("num".to_string(), "4".to_string())]),
+        )?;
         txn.commit()?;
     }
 
     {
         let mut txn = storage.txn()?;
         let mut tasks = txn.get_pending_tasks()?;
 
         let mut exp = vec![
```

djmitche (commented):
Really close! Just a few comments below.

taskchampion/src/replica.rs
Scandiravian: Thanks! Wouldn't have caught this myself

taskchampion/src/storage/inmemory.rs
Scandiravian: This is the part that I dislike because it's not readable enough

I'm not smart enough to figure out how to make it cleaner, so I've added a comment about what is going on to make it a bit more clear

taskchampion/src/storage/test.rs
Scandiravian: That was entirely my fault. Your original comment did suggest to use `JOIN`, but I guess I was a bit tired when I wrote it and I didn't make sure I understood the code-base well enough

In any case; thank you for really thinking through the impact of my coding choices, so I can avoid introducing a bug that would frustrate your users :raised_hands:

Scandiravian:
Thank you so much for the review and for taking the time to thoroughly explain the bug that my choice of `LEFT JOIN` would have caused - it helps me improve as a developer

I've added a commit for each comment you had to make it easier to read through. I hope it's okay to run the changes by you one final time before I squash (actually `fixup`) everything into a single commit so it's ready to merge :crossed_fingers:

taskchampion/src/storage/inmemory.rs
djmitche: That's probably good. I think this is actually pretty idiomatic Rust, but it does require a lot of context about the various types involved to understand it. The comment addresses that perfectly. 💯

taskchampion/src/storage/test.rs
```diff
@@ -273,18 +273,17 @@ pub(super) fn delete_task_exists(mut storage: impl Storage) -> Result<()> {
 }
 
 pub(super) fn get_pending_tasks(mut storage: impl Storage) -> Result<()> {
-    let uuid1 = Uuid::new_v4();
-    let uuid2 = Uuid::new_v4();
+    let uuid1 = Uuid::new_v4(); // in working set
+    let uuid2 = Uuid::new_v4(); // in working set
+    let uuid3 = Uuid::new_v4(); // in working set but does not exist
+    let uuid4 = Uuid::new_v4(); // not in working set but exists= Uuid::new_v4();
```
djmitche: Looks like a bit of extra text got included in that comment.

djmitche (approved):
Looks great, thanks! And no problem at all making another round of review.

I can squash this when merging, so if you want to just make one more commit to fix the comment, that's all that should be necessary.

taskchampion/src/storage/test.rs
Scandiravian: This is now fixed :smile:

Scandiravian:
Hey @djmitche

If everything looks good to you, feel free to squash and merge the PR :smile:

djmitche (approved):
Sorry for the delay!

Scandiravian:
> Sorry for the delay!

No problem. I wasn't sure if I had communicated clearly that I thought I had solved all remaining issues 😅

Thanks so much for all your constructive feedback on my changes. This was an incredibly pleasant experience and I hope I'll get the chance to contribute to the project again

Issue #457, opened 2024-09-19
Fix missing test coverage found by cargo-mutants


djmitche:
I heard about [cargo-mutants](https://mutants.rs/) a RustConf 2024, and thought I'd give it a go. It ran for about 2h and found the following. Maybe this is nonsense, but worth a look. It would run faster if I could convince it to only run the `lib` tests and not the integration tests.

   * [ ] taskchampion/src/server/cloud/gcp.rs:131:23: replace > with == in <impl Service for GcpService>::compare_and_swap in 5.5s build + 24.1s test                                                                                                                                                                          
   * [ ] taskchampion/src/taskdb/snapshot.rs:31:9: replace <impl Visitor for TaskDbVisitor>::expecting -> fmt::Result with Ok(Default::default()) in 5.6s build + 24.5s test                                                                                                                                                   
   * [x] taskchampion/src/storage/mod.rs:137:23: replace && with || in StorageTxn::is_empty in 5.2s build + 24.5s test                                                                                                                                                                                                         
   * [ ] taskchampion/src/server/cloud/gcp.rs:211:36: replace < with == in <impl Iterator for ObjectIterator<'a>>::next in 5.2s build + 24.8s test                                                                                                                                                                             
   * [ ] taskchampion/src/replica.rs:420:9: replace Replica::add_undo_point -> Result<()> with Ok(()) in 5.6s build + 32.2s test                                                                                                                                                                                               
   * [ ] taskchampion/src/task/task.rs:244:9: replace Task::get_legacy_udas -> impl Iterator<Item =(&str, &str)>+'_ with ::std::iter::once(("xyzzy", "xyzzy")) in 5.4s build + 25.1s test                                                                                                                                      
   * [ ] taskchampion/src/task/task.rs:252:9: replace Task::get_modified -> Option<Timestamp> with Some(Default::default()) in 5.3s build + 24.9s test                                                                                                                                                                         
   * [ ] taskchampion/src/server/sync/mod.rs:168:9: replace <impl Server for SyncServer>::add_snapshot -> Result<()> with Ok(()) in 5.3s build + 29.0s test                                                                                                                                                                    
   * [ ] taskchampion/src/taskdb/sync.rs:121:46: replace == with != in sync in 5.2s build + 24.1s test                                                                                                                                                                                                                         
[...]

PR #459, opened 2024-09-23
Additional tests for Replica methods

djmitche:
Address some missing test coverage identified by `cargo mutants` in #457.

### DIFF
--- a/taskchampion/src/replica.rs
+++ b/taskchampion/src/replica.rs
@@ -601,6 +601,11 @@ mod tests {
 
         // num_undo_points includes only the undo point
         assert_eq!(rep.num_undo_points().unwrap(), 1);
+
+        // A second undo point is counted.
+        let ops = vec![Operation::UndoPoint];
+        rep.commit_operations(ops).unwrap();
+        assert_eq!(rep.num_undo_points().unwrap(), 2);
     }
 
     #[test]
@@ -636,6 +641,13 @@ mod tests {
         assert_eq!(all_tasks.len(), 2);
         assert_eq!(all_tasks.get(&uuid1).unwrap().get_uuid(), uuid1);
         assert_eq!(all_tasks.get(&uuid2).unwrap().get_uuid(), uuid2);
+
+        let mut all_uuids = rep.all_task_uuids().unwrap();
+        all_uuids.sort();
+        let mut exp_uuids = vec![uuid1, uuid2];
+        exp_uuids.sort();
+        assert_eq!(all_uuids.len(), 2);
+        assert_eq!(all_uuids, exp_uuids);
     }
 
     #[test]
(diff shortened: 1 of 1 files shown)


taskchampion/src/replica.rs
```diff
@@ -601,6 +601,11 @@ mod tests {
 
         // num_undo_points includes only the undo point
         assert_eq!(rep.num_undo_points().unwrap(), 1);
+
+        // A second undo point is counted.
+        let ops = vec![Operation::UndoPoint];
+        rep.commit_operations(ops).unwrap();
+        assert_eq!(rep.num_undo_points().unwrap(), 2);
```
ryneeverett, line 608: I'd question the value of this. Looking at the cargo-mutants report:

> taskchampion/src/replica.rs:446:9: replace Replica::num_undo_points -> Result with Ok(1)

Is that a mistake we're actually likely to make?

taskchampion/src/replica.rs
```diff
@@ -636,6 +641,13 @@ mod tests {
         assert_eq!(all_tasks.len(), 2);
         assert_eq!(all_tasks.get(&uuid1).unwrap().get_uuid(), uuid1);
         assert_eq!(all_tasks.get(&uuid2).unwrap().get_uuid(), uuid2);
+
+        let mut all_uuids = rep.all_task_uuids().unwrap();
+        all_uuids.sort();
+        let mut exp_uuids = vec![uuid1, uuid2];
+        exp_uuids.sort();
+        assert_eq!(all_uuids.len(), 2);
+        assert_eq!(all_uuids, exp_uuids);
```
ryneeverett, line 650: Also seems like a questionable mutant:

> taskchampion/src/replica.rs:106:9: replace Replica::all_task_uuids -> Result<Vec> with Ok(vec![Default::default()])

taskchampion/src/replica.rs
```diff
@@ -759,6 +771,37 @@ mod tests {
...
+        assert!(!rep.commit_reversed_operations(ops)?);
+
+        // Commiting the correct operations succeeds
+        let ops = rep.get_undo_operations()?;
+        assert!(rep.commit_reversed_operations(ops)?);
+        assert_eq!(rep.num_undo_points().unwrap(), 1);
+
+        Ok(())
+    }
+
```
ryneeverett, line 804: In the spirit of fairness, this *does* seem like a valuable mutation to test against:

> taskchampion/src/replica.rs:355:12: delete ! in Replica::commit_reversed_operations

And a valuable test. :+1:

ryneeverett (approved):
On whole, the cargo-mutants tests seem to be a net improvement so i don't want to throw the brakes on this project. But I am questioning the value of testing against some of the individual mutants.

I guess the broader question is how far you intend to go with cargo-mutants -- if you're going to fix all the issues and add cargo-mutants to CI, then a few tests of marginal value are probably acceptable...

One thing I don't quite get is -- aren't there an infinite number of ways to screw up the code? I presume cargo-mutants only tries swapping "simple" values (for some definition of "simple").

djmitche:
> On whole, the cargo-mutants tests seem to be a net improvement so i don't want to throw the brakes on this project. But I am questioning the value of testing against some of the individual mutants.
>
> I guess the broader question is how far you intend to go with cargo-mutants -- if you're going to fix all the issues and add cargo-mutants to CI, then a few tests of marginal value are probably acceptable...

I'm not planning to add it to CI. But it did a pretty good job of finding some missing coverage, and it seems valuable to fix that. I'm not pursuing coverage for coverage's sake, nor planning on doing all of the checkboxes in #457. To be honest, adding tests like this is a pretty easy thing to do in the evening while watching TV, when I don't have the attention span to work on the operations project (#373).

> One thing I don't quite get is -- aren't there an infinite number of ways to screw up the code? I presume cargo-mutants only tries swapping "simple" values (for some definition of "simple").

I had this question, too, when I saw the talk at RustConf, but it was answered a few slides later: it has a set of a few tens of mutations that it applies, which through experience have been show to have good value. And that seems to be true here!

taskchampion/src/replica.rs
djmitche: No, but it did show that the tests weren't really exercising that function. That said, it's just deferring to the taskdb function of the same name and that is better tested, so I suppose this isn't very high-value.

taskchampion/src/replica.rs
djmitche: This particular miss just indicated to me that `all_task_uuids` wasn't covered by tests at all.

PR #465, opened 2024-10-05
Set up to use cargo-semver-checks on PRs and releases

### LINKED ISSUE #456: Set up cargo-semver-check to check for breaking changes

djmitche:
The check should read from the version in `taskchampion/Cargo.toml` so if that represents a major version bump then changes are allowed, as is the case here -- we have added and removed methods in the un-sealed trait `StorageTxn`. We should seal that, too.

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -101,3 +101,10 @@ jobs:
           command: fmt
           args: --all -- --check
 
+
+  semver-checks:
+    runs-on: ubuntu-latest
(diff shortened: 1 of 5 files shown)


djmitche:
The [docs for these types](https://docs.rs/taskchampion/latest/taskchampion/) say

> An implementation is provided, but users of this crate can provide their own implementation as well.

Sealing would prevent that, so they are correct as they are (not sealed).

ryneeverett (approved):

PR #466, opened 2024-10-12
Revert incomplete "expose operations" work

djmitche:
This reverts all of the PRs made to support #373 so far, with the intent of trying again after the 0.8.0 release, and when the solution is better understood.


ryneeverett (approved):
I don't have a problem with this, but do you @djmitche think this is the best course of action? I don't want to win if it's my opinion against yours.

It does seem a shame to lose PR's like #453 and #459 which were self-contained improvements.

djmitche:
I think it's a reasonable course of action. I'll re-land this on a branch and continue working on it, and if that ultimately is rejected, we can pull out those two bits and land them separately.

PR #468, opened 2024-10-13
Allow using native TLS root certs

### LINKED ISSUE #460: Use system CA roots
From https://github.com/GothenburgBitFactory/taskwarrior/issues/3625, TC should have the option of using the system CA roots for validating connections to taskchampion-sync-server.

I don't think this should be the default, as it makes MITM attacks easier (in fact that's the point, allowing corporate IT to MITM connections). But, it should be an option.

Let's add a `tls-native-roots` feature that instructs `reqwest` to use its `rustls-tls-native-roots` feature, similar to https://github.com/GothenburgBitFactory/taskwarrior/pull/3626.

djmitche:
Fixes #460.

### DIFF
--- a/.github/workflows/rust-tests.yml
+++ b/.github/workflows/rust-tests.yml
@@ -16,6 +16,7 @@ jobs:
         features:
           - ""
           - "server-sync"
+          - "tls-native-roots"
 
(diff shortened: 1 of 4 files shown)


djmitche:
(I think the sqlite3 change was from a recent update to the GitHub action image)

ryneeverett (approved):

PR #470, opened 2024-10-13
Use lib docstring as crates.io README

djmitche:
This makes https://crates.io/crates/taskchampion and https://docs.rs/taskchampion/ display (roughly) the same information. The symbol links like ``[`Replica`]`` won't work on on crates.io but I think that's OK (I've certainly seen it on other crates).


ryneeverett (approved):
Seems like a good tradeoff to me. I'm not sure I'd ever even looked at the crates.io website before just now (for any project, let alone taskchampion).

PR #474, opened 2024-10-25
Expose operations, including "historical" operations

### LINKED ISSUE #373: Expose operations from TaskChampion

djmitche:
This stores all operations for all existing tasks, including operations that were entered on other replicas, and makes those available via `get_task_operations`. This functionality will support `task info` in TaskWarrior, and can support similar "task history" displays in other consumers.

As the last commit and the docstrings describe, this isn't perfect -- if there are conflicting operations, then the results may differ between replicas. But those are rare, and in practice this is sufficient to provide information to users. In particular, users who never `task sync` will see correct information.

This is a large diff, but broken up into commits, many of which were reviewed and landed previously, but backed out for the 0.8.0 release.  Feel free to review commit-by-commit.

Fixes #373.


taskchampion/src/operation.rs
```diff
@@ -55,6 +56,85 @@ impl Operation {
...
+                        },
+                    ) => Equal
+                        // Sort Updates principally by timestamp.
+                        .then(uuid1.cmp(uuid2))
+                        .then(timestamp1.cmp(timestamp2))
+                        .then(property1.cmp(property2))
+                        .then(value1.cmp(value2))
+                        .then(old_value1.cmp(old_value2)),
+                    (UndoPoint, UndoPoint) => Equal,
+                    // a and b are always the same variant.
```
ryneeverett: "a and b"? Does this comment refer to the default/unreachable case? I don't understand it.

taskchampion/src/operation.rs
```diff
@@ -55,6 +56,85 @@ impl Operation {
...
+    /// multiple creations of the same task.
+    fn cmp(&self, other: &Self) -> Ordering {
+        use Operation::*;
+        use Ordering::*;
+        fn type_idx(op: &Operation) -> u8 {
+            match op {
+                Create { .. } => 1,
+                Update { .. } => 2,
+                Delete { .. } => 3,
+                UndoPoint => 4,
```
ryneeverett: I understand that UndoPoint's come first because they have no uuid but it seems a little too clever and potentially confusing to make them "4". Maybe `u8::MAX` to make it more clearly a throwaway value? An explanatory comment wouldn't hurt.

taskchampion/src/taskdb/sync.rs
```diff
@@ -129,6 +139,12 @@ pub(super) fn sync(
         }
     }
 
+    // Add the transformed server ops to the DB. Critically, these are immediately marked as synced
+    // and thus not subject to any of the invariants around operations and task state.
+    for o in transformed_server_ops {
+        txn.add_operation(dbg!(o).into_op())?;
```
ryneeverett: dbg!

taskchampion/src/taskdb/sync.rs
```diff
@@ -129,6 +139,12 @@ pub(super) fn sync(
         }
     }
 
+    // Add the transformed server ops to the DB. Critically, these are immediately marked as synced
```
ryneeverett, line 142: "Critically, these are immediately marked as synced" -- Where does this marking happen? (Underlying question: How are we confident that this critical assertion is true?) In the inmemory implementation it appears to be the opposite (`operations.push((false, op));`).

taskchampion/src/taskdb/sync.rs
```diff
@@ -278,6 +305,50 @@ mod test {
         sync(&mut server, db1.storage.txn()?.as_mut(), false).unwrap();
         assert_eq!(db1.sorted_tasks(), db2.sorted_tasks());
 
+        for (dbnum, db) in [(1, &mut db1), (2, &mut db2)] {
+            dbg!(dbnum);
```
ryneeverett: dbg! (I presume one would never intentionally commit a dbg! even in a test?)

taskchampion/src/replica.rs
```diff
@@ -796,6 +823,37 @@ mod tests {
...
+        let uuid3 = Uuid::new_v4();
+
+        let mut rep = Replica::new_inmemory();
+
+        let mut ops = Operations::new();
+        ops.push(Operation::UndoPoint);
+        rep.create_task(uuid1, &mut ops).unwrap();
+        ops.push(Operation::UndoPoint);
+        rep.create_task(uuid2, &mut ops).unwrap();
+        rep.commit_operations(dbg!(ops))?;
```
ryneeverett: dbg!

taskchampion/src/storage/sqlite.rs
```diff
@@ -107,17 +107,11 @@ impl SqliteStorage {
         for q in create_tables {
             con.execute(q, []).context("Creating table")?;
         }
-        // At this point the DB schema is that of TaskChampion 0.7.0.
+
+        // -- At this point the DB schema is that of TaskChampion 0.7.0.
```
ryneeverett: Should all the "0.7.0" version references (also appearing as "0_7_0") be bumped to "0.8.0"?

ryneeverett (approved):

taskchampion/src/taskdb/sync.rs
djmitche: The `sync_complete` call just below does so. I'll clairfy.

ryneeverett: Oh, that makes perfect sense. As I figured, there is a simple answer and I just wasn't making the connection.

djmitche:
I added commits for some of the comments, and updated existing commits for the minor bits (dbg!, mostly)

ryneeverett (approved):

PR #476, opened 2024-10-27
Update sync protocol docs

djmitche:
This
 * clarifies when _gone_ and _not-found_ responses should be sent
 * includes `X-Parent-Version-Id` in the `GetChildVersion` response, since taskchampion-sync-server is already sending that value.

See GothenburgBitFactory/taskchampion-sync-server#47.

### DIFF
--- a/docs/src/sync-protocol.md
+++ b/docs/src/sync-protocol.md
@@ -95,7 +95,7 @@ If found, it returns the version's
 If found, it returns a _not-found_ indication.
 Note that this circumstance is not an error, and occurs during every successful sync process.
 
-Implementations may return a different error code if the version is "gone": not found _and would not be accepted by AddVersion_, in order to provide better error messages to the user.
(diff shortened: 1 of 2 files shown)


docs/src/sync-protocol.md
```diff
@@ -92,9 +92,15 @@ If found, it returns the version's
  * parent version ID (matching that in the request), and
  * encrypted version data.
 
-If not found, it returns an indication that no such version exists.
+If found, it returns a _not-found_ indication.
 Note that this circumstance is not an error, and occurs during every successful sync process.
 
+Implementations may return a different error code if the version is "gone": not found _and would not be accepted by AddVersion_, in order to provide better error messages to the user.
```
ryneeverett:
```suggestion
Implementations may return a different error code if the version is "gone" (not found _and would not be accepted by AddVersion_) in order to provide better error messages to the user.
```

ryneeverett (approved):

PR #485, opened 2024-11-27
Record the DB version in the DB itself.

### LINKED ISSUE #477: Record DB versions in the DB itself, and refuse to downgrade
0.9.0 added some schema upgrades, and that might cause issues on a downgrade. Actually, the schemas are compatible, but 0.9.0 leaves sync'd operations in the operations table, using a new boolean column to distinguish synced and un-synced. 0.8.0 doesn't know to look for that column.

For the moment, the 0.9.0 release warns about this, but going forward let's store the TC version in the DB, and just refuse to run with a DB that's newer than the current version of TC.  That should both make upgrades easier to handle (`if db_version < latest { .. }`) and prevent issues with downgrades.

If users do need to downgrade, then deleting the DB and re-syncing should do the job for them.

djmitche:
The version is not the same as the TaskChampion version, but uses a major/minor scheme to represent compatibility between versions.

Fixes #477.

### DIFF
--- a/taskchampion/src/storage/sqlite/schema.rs
+++ b/taskchampion/src/storage/sqlite/schema.rs
@@ -1,0 +1,236 @@
+use crate::errors::{Error, Result};
+use anyhow::Context;
+use rusqlite::{params, Connection, OptionalExtension, Transaction};
+
+/// A database schema version.
+///
+/// The first value is the major version, with different major versions completely incompatible
+/// with one another.
+///
+/// The second is the minor version, with all minor versions in the same major version being
+/// compatible with one another. That is, a TaskChampion binary with a latest version of (MAJ, MIN)
+/// can safely operate on any DB with major version MAJ, whether its minor version is greater or
+/// smaller than MIN, and can upgrade that DB to (MAJ, MIN) if necessary.
+///
+/// For example, a new index would trigger an increment of the minor version, as older versions of
+/// TaskChampion can still safely use the DB with the addition of the index.
+#[derive(Copy, Clone, Debug, PartialOrd, Ord, PartialEq, Eq)]
+pub(super) struct DbVersion(pub u32, pub u32);
+
+type UpgradeFn = fn(&Transaction) -> Result<()>;
+
+/// DB Versions and functions to upgrade to them.
+///
+/// Add new vesions here, in order. Prefer minor version bumps for better compatibility, using
+/// techniques like virtual columns where possible.
+const VERSIONS: &[(DbVersion, UpgradeFn)] = &[(DbVersion(0, 1), upgrade_to_0_1)];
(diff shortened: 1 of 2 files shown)


taskchampion/src/storage/sqlite/schema.rs
```diff
@@ -0,0 +1,209 @@
...
+/// that is considered version (0, 0).
+pub(super) fn get_db_version(con: &Connection) -> Result<DbVersion> {
+    let version: Option<(u32, u32)> = match con
+        .query_row("SELECT major, minor FROM version", [], |r| {
+            Ok((r.get("major")?, r.get("minor")?))
+        })
+        .optional()
+    {
+        Ok(v) => v,
+        Err(rusqlite::Error::SqliteFailure(_, _)) => None,
```
ryneeverett: Shouldn't a SqliteFailure log a warning?

taskchampion/src/storage/sqlite/schema.rs
```diff
@@ -0,0 +1,209 @@
...
+        Err(e) => return Err(dbg!(e).into()),
```
ryneeverett: Prefer `log::debug!`?

taskchampion/src/storage/sqlite/schema.rs
```diff
@@ -0,0 +1,209 @@
...
+        let t = con.transaction()?;
+        assert!(has_column(&t, "operations", "id")?);
+        assert!(has_column(&t, "operations", "data")?);
+        assert!(has_column(&t, "operations", "uuid")?);
+        assert!(has_column(&t, "sync_meta", "key")?);
+        assert!(has_column(&t, "sync_meta", "value")?);
+        assert!(has_column(&t, "tasks", "uuid")?);
+        assert!(has_column(&t, "tasks", "data")?);
+        assert!(has_column(&t, "working_set", "id")?);
+        assert!(has_column(&t, "working_set", "uuid")?);
```
ryneeverett: Perhaps throw in a `assert_eq!(get_db_version(&con)?, DbVersion(0, 1));`?

ryneeverett (approved):

taskchampion/src/storage/sqlite/schema.rs
djmitche: Oops, that was left in accidentally!

taskchampion/src/storage/sqlite/schema.rs
djmitche: Hm, this is intended to catch a missing table, but we could possibly check that and fail with an error if the table does exist.

djmitche: I've done that, but I can't see a good way to put a Connection into an error state as part of tests. Opening a DB at a non-existent path fails in `open`. Somehow corrupting the DB after opening it seems too fragile a thing to do in a test.

PR #492, opened 2024-12-01
Add an AWS sync backend

### LINKED ISSUE #368: Sync to AWS
Similar to the GCP sync implemented in GothenburgBitFactory/taskwarrior#3185, we should be able to sync replicas to AWS's object storage.

The tricky bit here is that, unlike GCS and Azure Blob, S3 does not provide a compare-and-swap operation.

Some reading suggests that the easiest way to accomplish this would be to use DynamoDB as a lock over the "latest" object in the S3 bucket.

djmitche:
This approach is very similar to that of GCP, and utilizes a similar conditional-write mechanism to prevent conflicting updates to the `latest` key.                                                                                                                                                                                                                                                                                                                  
                                                                                                                                                                                                                                                                                                                               
This updates the package major version to 1.0.0, as it modifies the `ServerConfig` enum by adding a new variant -- a semver-breaking change.  To avoid further major version bumps from additional variants, this also adds `#[non-exhaustive]` to the enum.                                                                                                                                                                                                                                                                                          

This also bumps the MSRV to the MSRV of the AWS SDK (1.78.0).

### DIFF
--- a/taskchampion/src/server/cloud/aws.rs
+++ b/taskchampion/src/server/cloud/aws.rs
@@ -1,0 +1,631 @@
+use super::service::{ObjectInfo, Service};
+use crate::errors::Result;
+use aws_config::{
+    meta::region::RegionProviderChain, profile::ProfileFileCredentialsProvider, BehaviorVersion,
+    Region,
+};
+use aws_credential_types::Credentials;
+use aws_sdk_s3::{
+    self as s3,
+    error::ProvideErrorMetadata,
+    operation::{get_object::GetObjectOutput, list_objects_v2::ListObjectsV2Output},
+};
+use std::future::Future;
+use tokio::runtime::Runtime;
+
+/// A [`Service`] implementation based on the Google Cloud Storage service.
+pub(in crate::server) struct AwsService {
+    client: s3::Client,
+    rt: Runtime,
+    bucket: String,
+}
+
+/// Credential configuration for access to the AWS service.
+///
+/// These credentials must have a least the following policy, with BUCKETNAME replaced by
+/// the bucket name:
+///
+/// ```json
+/// {
+///     "Version": "2012-10-17",
+///     "Statement": [
+///         {
+///             "Sid": "TaskChampion",
+///             "Effect": "Allow",
+///             "Action": [
+///                 "s3:PutObject",
+///                 "s3:GetObject",
+///                 "s3:ListBucket",
+///                 "s3:DeleteObject"
+///             ],
+///             "Resource": [
+///                 "arn:aws:s3:::BUCKETNAME",
+///                 "arn:aws:s3:::BUCKETNAME/*"
+///             ]
+///         }
+///     ]
+/// }
+/// ```
+#[non_exhaustive]
+pub enum AwsCredentials {
+    /// A pair of access key ID and secret access key.
+    AccessKey {
+        access_key_id: String,
+        secret_access_key: String,
+    },
+    /// A named profile from the profile files in the user's home directory.
+    Profile { profile_name: String },
+    /// Use the [default credential
+    /// sources](https://docs.rs/aws-config/latest/aws_config/default_provider/credentials/struct.DefaultCredentialsChain.html),
+    /// such as enviroment variables, the default profile, or the task/instance IAM role.
+    Default,
+}
+
+impl AwsService {
+    pub(in crate::server) fn new(
+        region: String,
+        bucket: String,
+        creds: AwsCredentials,
+    ) -> Result<Self> {
+        let rt = Runtime::new()?;
+
+        let config =
+            rt.block_on(async {
+                let mut config_provider = aws_config::defaults(BehaviorVersion::v2024_03_28());
+                match creds {
+                    AwsCredentials::AccessKey {
+                        access_key_id,
+                        secret_access_key,
+                    } => {
+                        config_provider = config_provider.credentials_provider(
+                            Credentials::from_keys(access_key_id, secret_access_key, None),
+                        );
+                    }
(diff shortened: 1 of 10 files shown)


djmitche:
@dathanb @travisb please take a look?

travisb:
@djmitche p sure you meant to tag someone else

djmitche:
Sorry -- @travisby!

taskchampion/src/server/config.rs
```diff
@@ -65,6 +71,22 @@ pub enum ServerConfig {
         /// be any suitably un-guessable string of bytes.
         encryption_secret: Vec<u8>,
     },
+    /// An Amazon Web Servicesstorage bucket.
```
ryneeverett:
```suggestion
    /// An Amazon Web Services storage bucket.
```

taskchampion/src/server/cloud/aws.rs
```diff
@@ -0,0 +1,628 @@
+#![allow(unreachable_code, unused_variables)]
```
ryneeverett: Can these can be deleted now?

taskchampion/src/server/cloud/aws.rs
```diff
@@ -0,0 +1,628 @@
...
+///             ],
+///             "Resource": [
+///                 "arn:aws:s3:::BUCKETNAME",
+///                 "arn:aws:s3:::BUCKETNAME/*"
+///             ]
+///         }
+///     ]
+/// }
+/// ```
+#[non_exhaustive]
```
ryneeverett, line 49: Should this be non_exhaustive? My understanding is that the primary motivation for non_exhaustive enums is to force outside crates to match the wildcard because you expect to add variants in the future which you don't want to break those outside crates. So I totally get why ServerConfig should be non_exhaustive -- we'll probably add more server backends and configuration options -- but I don't get it for AwsCredentials -- do we expect to add more variants here?

taskchampion/src/server/cloud/aws.rs
```diff
@@ -0,0 +1,628 @@
...
+        self.rt.block_on(fut)
+    }
+}
+
+/// Convert an object name from bytes to a string.
+fn name_to_string(name: &[u8]) -> String {
+    String::from_utf8(name.to_vec()).expect("non-UTF8 object name")
+}
+
+/// Convert an error that can be converted to `s3::Error` (but not [`crate::Error`] into
```
ryneeverett:
```suggestion
/// Convert an error that can be converted to `s3::Error` (but not [`crate::Error`]) into
```

taskchampion/src/server/cloud/aws.rs
```diff
@@ -0,0 +1,628 @@
...
+    }
+
+    #[test]
+    fn compare_and_swap_matches() {
+        let Some((mut svc, pfx)) = make_service() else {
+            return;
+        };
+
+        // Create the existing file, with two generations.
+        svc.put(&pfx("testy"), b"foo1").unwrap();
```
ryneeverett: I don't understand your goal with the "two generations" here. How might the original write of that file impact the assertions?

taskchampion/src/server/cloud/aws.rs
```diff
@@ -0,0 +1,628 @@
...
+    }
+
+    #[test]
+    fn compare_and_swap_mismatch() {
+        let Some((mut svc, pfx)) = make_service() else {
+            return;
+        };
+
+        // Create the existing file, with two generations.
+        svc.put(&pfx("testy"), b"foo1").unwrap();
```
ryneeverett: Again, I do not understand the purpose of the original `put`. (Assuming we're only trying to test taskchampion here, and not the aws library or service.)

taskchampion/src/server/cloud/aws.rs
```diff
@@ -0,0 +1,628 @@
...
+        };
+
+        // Create the existing file, but since it is named "racing-put" the file will appear just
+        // before the `put_object` call.
+        assert!(!svc
+            .compare_and_swap(&pfx("racing-put"), None, b"bar".to_vec())
+            .unwrap());
+        let got = svc.get(&pfx("racing-put")).unwrap();
+        assert_eq!(got, Some(b"CHANGED".to_vec()));
+    }
```
ryneeverett: I don't understand the name of this test and it isn't clear to me how the behavior tested is distinct from that of `compare_and_swap_changes`.

ryneeverett (approved):

taskchampion/src/server/cloud/aws.rs
djmitche: Your understanding is correct. Here, we might want to add more variants (AWS has a lot of credential options!). I don't expect anything outside this crate to use `match` on this type, but for the sake of semver it's still worthwhile to reserve the possibility of extending this later.

To be honest, if I could go back in time I'd suggest that `pub enum` be non-exhaustive by default.

taskchampion/src/server/cloud/aws.rs
djmitche: I copied this from the gcp tests, and I don't remember if there was a failure mode I had in mind. I think the more important thing is that an `existing_value` that matches a non-current value in the object store should cause `compare_and_swap` to fail. So I'll assert that as well.

djmitche: Oh, now that I look at your next comment, `compare_and_swap_mismatch` is checking that, so I'll leave the two tests with some better comments.

taskchampion/src/server/cloud/aws.rs
djmitche: I've added some comments.

PR #497, opened 2024-12-06
Use strings for cloud service object names

### LINKED ISSUE #489: Modify the Service trait to operate in terms of strings
It's currently using `&[u8]`, but both GCP and AWS (#368) implementations convert those to strings and panic if not UTF-8.

djmitche:
Fixes #489.


taskchampion/src/server/cloud/server.rs
```diff
@@ -545,11 +545,7 @@ mod tests {
     impl std::fmt::Debug for MockService {
         fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
             f.debug_map()
-                .entries(
-                    self.0
-                        .iter()
-                        .map(|(k, v)| (std::str::from_utf8(k).unwrap(), v)),
-                )
+                .entries(self.0.iter().map(|(k, v)| (k, v)))
```
ryneeverett: Surely this `map` is superfluous now? Perhaps the whole `impl`?

taskchampion/src/server/cloud/service.rs
```diff
@@ -16,23 +16,34 @@ pub(in crate::server) struct ObjectInfo {
...
-        name: &[u8],
+        name: &str,
         existing_value: Option<Vec<u8>>,
         new_value: Vec<u8>,
     ) -> Result<bool>;
 }
+
+/// Enforce the limits on `name` described for [`Service`].
+///
+/// Names are generated by `server.rs` in a way that is not affected by user intpu, so this only
```
ryneeverett:
```suggestion
/// Names are generated by `server.rs` in a way that is not affected by user input, so this only
```

ryneeverett (approved):

taskchampion/src/server/cloud/server.rs
djmitche: The map is! The impl isn't used but is useful for dbg!ging. Since it's test-only, I don't think there's harm in keeping it.

djmitche: The map is! The impl isn't used but is useful for dbg!ging. Since it's test-only, I don't think there's harm in keeping it.

PR #506, opened 2024-12-11
Make AwsCredentials type public

djmitche:
I went to look at https://docs.rs/taskchampion/latest/taskchampion/server/enum.ServerConfig.html and this type is private. Oops :(


djmitche:
Required for https://github.com/GothenburgBitFactory/taskwarrior/issues/3721

ryneeverett (approved):

PR #510, opened 2024-12-21
Refactor Expire Tasks

### LINKED ISSUE #436: Use `all_task_data` in `Replica::expire_tasks`
(This will make sense when #372 is finished)

`Replica::expire_tasks` currently uses `Task` instead of `TaskData`, meaning that it is doing extra work to load task dependencies. Refactor it to use just `TaskData` objects, parsing the `"modified"` property and comparing it to the expiration time.

zaxioms0:
I wrote a small patch to address #436. This is my first PR so I'm not sure if I did it correctly, but it seems to pass all of the tests?

### DIFF
--- a/src/replica.rs
+++ b/src/replica.rs
@@ -8,10 +8,11 @@ use crate::taskdb::TaskDb;
 use crate::workingset::WorkingSet;
 use crate::{Error, TaskData};
 use anyhow::Context;
-use chrono::{Duration, Utc};
+use chrono::{DateTime, Duration, Utc};
 use log::trace;
 use std::collections::HashMap;
 use std::rc::Rc;
+use std::time::UNIX_EPOCH;
 use uuid::Uuid;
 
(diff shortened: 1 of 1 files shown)


zaxioms0:
I PR'ed the wrong clone... 🤦

djmitche (approved):
Nice work! Thanks!

src/replica.rs
```diff
@@ -422,20 +422,18 @@ impl Replica {
     pub fn expire_tasks(&mut self) -> Result<()> {
         let six_mos_ago = Utc::now() - Duration::days(180);
         let mut ops = Operations::new();
-        self.all_tasks()?
+        self.all_task_data()?
             .drain()
-            .filter(|(_, t)| t.get_status() == Status::Deleted)
```
djmitche, line 427: Oh no -- this line should not have been removed! We're now expiring _all_ old tasks, whether deleted or not!

PR #511, opened 2024-12-24
Introduce an access mode for SQLite storage

### LINKED ISSUE #479: Support opening Replica in read-only mode
Lots of access to the Replica is only reading, not writing. SQLite can be quite a bit faster in that circumstance. Also, SQLite writes to the database itself when opened in read-write mode (https://github.com/GothenburgBitFactory/taskwarrior/issues/3418).

Let's allow opening a Replica in read-only mode.

djmitche:
The storage implementation checks this access mode for all methods, but also opens SQLite in read-only mode when appropriate, which causes SQLite to skip creating a WAL log or updating the access time in the database.

As a concession to ease-of-use with a new task database, the database is opened in read-write mode and the schema set up if it does not exist, regardless of the passed `AccessMode`.

This is a breaking change because:

 - The publicly-visible method `SqliteStorage::new` now takes a
   different number of parameters.
 - The public enum `StorageConfig` has been marked #[non_exhaustive].
   Pattern-matching on it outside of its crate must now include a
   wildcard pattern like `_`, or it will fail to compile.
 - `StorageConfig::OnDisk` has an additional field, `access_mode`, which
   has to be included when constructing or matching on this variant.

[edited to add the breaking-change justification]

### DIFF
--- a/src/storage/config.rs
+++ b/src/storage/config.rs
@@ -2,15 +2,26 @@ use super::{InMemoryStorage, SqliteStorage, Storage};
 use crate::errors::Result;
 use std::path::PathBuf;
 
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum AccessMode {
+    ReadOnly,
+    ReadWrite,
+}
+
 /// The configuration required for a replica's storage.
+#[non_exhaustive]
 pub enum StorageConfig {
     /// Store the data on disk.  This is the common choice.
     OnDisk {
         /// Path containing the task DB.
         taskdb_dir: PathBuf,
 
-        /// Create the DB if it does not already exist
+        /// Create the DB if it does not already exist. This will occur
+        /// even if access_mode is `ReadOnly`.
         create_if_missing: bool,
+
+        /// Access mode for this database.
+        access_mode: AccessMode,
     },
     /// Store the data in memory.  This is only useful for testing.
     InMemory,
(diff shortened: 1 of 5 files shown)


djmitche:
I'm not sure about deprecating the `StorageConfig` thing -- it seems like I'm designing around semver. And, we still have `ServerConfig` following this model.

I'm open to other ideas for how to add a read-only mode in a backward-compatible fashion, or suggestions to just go to 2.0.0.

djmitche:
>  taskchampion::storage::SqliteStorage::new now takes 3 parameters instead of 2, in /home/runner/work/taskchampion/taskchampion/src/storage/sqlite.rs:86

Oops, I guess even this is a breaking change. That type probably shouldn't have been exposed to begin with!

Maybe I should take a hard look at the API and make a 2.0.0 with a bunch of API fixes.

src/storage/config.rs
```diff
@@ -2,27 +2,44 @@ use super::{InMemoryStorage, SqliteStorage, Storage};
...
         create_if_missing: bool,
+
+        /// Access mode for this database.
+        access_mode: AccessMode,
     },
     /// Store the data in memory.  This is only useful for testing.
     InMemory,
 }
 
+#[allow(deprecated)]
```
ryneeverett: Is this needed? From your PR comments I'm under the impression you were going to deprecate something and then changed your mind.

ryneeverett (approved):

src/storage/config.rs
djmitche: Oh, thanks, I missed this one!

PR #526, opened 2025-01-05
Make AccessMode type public

djmitche:
I did it again:

image

That's a private type (here seen as not linkified) in a public type. This is why we had 1.0.2 as well. I tried enabling the `unreachable_pub` lint, but it didn't help -- I [think that's a bug](https://github.com/rust-lang/rust/issues/135134).

### DIFF
--- a/src/errors.rs
+++ b/src/errors.rs
@@ -66,7 +66,7 @@ impl From<ureq::Error> for Error {
     }
 }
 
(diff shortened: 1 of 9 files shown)


djmitche:
In that linked bug @Urgau pointed out that I want a different link, `unnameable_types`, and indeed that does detect this bug.

ryneeverett (approved):

PR #542, opened 2025-01-28
Include examples of operations handling in docs

### LINKED ISSUE #539: Add examples of operation manipulation to docs
https://docs.rs/taskchampion/latest/taskchampion/struct.Replica.html and maybe the top level should have some examples of handling operations, since that's an unusual API.

djmitche:
Fixes #539.

### DIFF
--- a/src/crate-doc.md
+++ b/src/crate-doc.md
@@ -29,7 +29,36 @@ Replica synchronization takes place against a server.
 Create a server with [`ServerConfig`].
 
 The [`server`] module defines the interface a server must meet.
-Users can define their own server impelementations.
+Several server implementations are included, and users can define their own implementations.
+
+# Example
+
+```rust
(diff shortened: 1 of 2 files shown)


ryneeverett (commented):
Your editor seems to be confused about filetypes. Pretty sure you wanted `///` in the rust file rather than `#` or `/** ... **/` and pretty sure you wanted neither in the markdown file.

djmitche:
The `# ` omits those lines from the generated HTML. This accomplishes showing the setup and sync at the top level doc, and the operations in the Replica doc, while still allowing each to run as a test case.

I used `/** .. **/` instead of `///` so it's easier to reformat the example text. It renders to HTML just fine.

IOW, I think this is correct!

ryneeverett (approved):
Thanks for the explanation! That makes perfect sense and also explains why the indentation is the way it is.

Issue #540, opened 2025-01-27
Proposal: api change around user defined attributes


grasegger:
The idea came to me based on this little chat history https://discord.com/channels/796949983734661191/796964492235767858/1333079601941057536 (+4 messages)

> karl:  so the taskchampion book mentions legacy uda which do not follow <namespace>.<key>, but is there actually a way to define a "modern" uda in the config right now?
> djmitche : we should probably change that
> djmitche: Taskwarrior only uses "legacy" UDAs
> djmitche: although it's encoded in the API so shrug
> djmitche: (that was from the time when TaskChampion was "inspired by" Taskwarrior)

I would propose the following changes: 
- [x] 1) Introduce new functions that don't speak about the quality of user defined attributes that are used within taskwarrior
```
Task::get_user_defined_attribute
Task::set_user_defined_attribute
Task::get_user_defined_attributes
Task::remove_user_defined_attribute
```
- [x] 2) Deprecate the following functions:
```
taskchampion::Task::get_uda
taskchampion::Task::set_uda
taskchampion::Task::get_udas
taskchampion::Task::remove_uda
taskchampion::Task::get_legacy_uda
taskchampion::Task::set_legacy_uda
taskchampion::Task::get_legacy_udas
taskchampion::Task::remove_legacy_uda
``` 
- [x]  3) Make legacy functions a wrapper around the new functions (so they can be deleted in a future breaking release)
- [x]  3) Update the taskchampion book to explain that legcay udas will continue to be first class citizens and thus not be called legacy going forward. Rather namespaced udas are being deprecated.
- [x]  4) Update the function docs to tell developers why this is deprecated

If this is ok with the maintainers I would be willing to take on this task.

djmitche:
Sounds like a good idea!

linuxcaffe:
Well what the hell? If you intend to eventually break everything including UDAs, how/ why are you still calling it taskwarrior? And you want to eventually lure every task user to one way "upgrade"? To what end? So you can finally say you killed taskwarrior? You're not even really a task user, don't understand or appreciate half the features you intend to break, and you don't give a crap about the countess task users who have done their own amazing custom shit (mostly using UDAs) 

Go ahead reinvent taskwarrior, co-opt any code you fancy, deprecate till your heart's content, but call it Taskchapion, stop calling it TW3 (cause it ain't, it's a trap) and stop trying to sabotage a well loved program, for the ego boost!

[mic drop]

[...]

PR #545, opened 2025-01-31
Deprecated namespaced UDAs; promote legacy UDAs to be default (#540)

grasegger:
This commit:
- Adds documentation about why namespaced UDAs existed and how they are an old idea
- Deprecates functions for namespaced and legacy UDAs
- Adds functions for handling taskwarrior compatible UDAs instead of relying on functions using the `legacy` word

About the deprecation of namespaced UDAs:

When TaskChampion was first implemented there was an idea to only use
namespaced keys in the future. This was before TaskChampion was integrated into
Taskwarrior 3, but it lead to UDAs that are currently being used being referenced
as "legacy" UDAs.

This commit changes that idea: Namespaced UDAs are deprecated, and what was called
legacy will be the way for UDAs going forward.

### DIFF
--- a/docs/src/tasks.md
+++ b/docs/src/tasks.md
@@ -48,7 +48,3 @@ Any unrecognized keys are treated as "user-defined attributes" (UDAs).
 These attributes can be used to store additional data associated with a task.
 For example, applications that synchronize tasks with other systems such as calendars or team planning services might store unique identifiers for those systems as UDAs.
 The application defining a UDA defines the format of the value.
-
-UDAs _should_ have a namespaced structure of the form `<namespace>.<key>`, where `<namespace>` identifies the application defining the UDA.
-For example, a service named "DevSync" synchronizing tasks from GitHub might use UDAs like `devsync.github.issue-id`.
-Note that many existing UDAs for Taskwarrior integrations do not follow this pattern; these are referred to as legacy UDAs.
--- a/src/task/task.rs
+++ b/src/task/task.rs
@@ -219,13 +219,16 @@ impl Task {
     /// Get the named user defined attributes (UDA).  This will return None
     /// for any key defined in the Task data model, regardless of whether
     /// it is set or not.
+    #[deprecated(note = "namespaced UDAs will not be supported in the future")]
     pub fn get_uda(&self, namespace: &str, key: &str) -> Option<&str> {
+        #[allow(deprecated)]
         self.get_legacy_uda(uda_tuple_to_string(namespace, key).as_ref())
     }
 
     /// Get the user defined attributes (UDAs) of this task, in arbitrary order.  Each key is split
     /// on the first `.` character.  Legacy keys that do not contain `.` are represented as `("",
     /// key)`.
+    #[deprecated(note = "namespaced UDAs will not be supported in the future")]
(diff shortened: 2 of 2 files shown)


docs/src/tasks.md
```diff
@@ -49,6 +49,13 @@ These attributes can be used to store additional data associated with a task.
 For example, applications that synchronize tasks with other systems such as calendars or team planning services might store unique identifiers for those systems as UDAs.
 The application defining a UDA defines the format of the value.
 
-UDAs _should_ have a namespaced structure of the form `<namespace>.<key>`, where `<namespace>` identifies the application defining the UDA.
-For example, a service named "DevSync" synchronizing tasks from GitHub might use UDAs like `devsync.github.issue-id`.
-Note that many existing UDAs for Taskwarrior integrations do not follow this pattern; these are referred to as legacy UDAs.
+#### Namespaced UDAs
+
+When TaskChampion was first implemented there was an idea to only use namespaced
+keys in the future. This was before TaskChampion became Taskwarrior 3, but it
```
ryneeverett:
```suggestion
keys in the future. This was before TaskChampion was integrated into Taskwarrior 3, but it
```

docs/src/tasks.md
ryneeverett, line 1: Would it make more sense to just remove this paragraph than to expand upon a deprecated concept?

src/task/task.rs
```diff
@@ -235,15 +238,30 @@ impl Task {
 
     /// Get the named user defined attribute (UDA) in a legacy format.  This will return None for
     /// any key defined in the Task data model, regardless of whether it is set or not.
+    #[deprecated(note = "please use Task::get_user_defined_attribute")]
     pub fn get_legacy_uda(&self, key: &str) -> Option<&str> {
+        self.get_user_defined_attribute(key)
+    }
+
+    /// Get the named user defined attribute (UDA) This will return None for any key
```
ryneeverett:
```suggestion
    /// Get the named user defined attribute (UDA). This will return None for any key
```

src/task/task.rs
```diff
@@ -235,15 +238,30 @@ impl Task {
...
+    pub fn get_user_defined_attribute(&self, key: &str) -> Option<&str> {
         if Task::is_known_key(key) {
             return None;
         }
         self.data.get(key)
     }
 
     /// Like `get_udas`, but returning each UDA key as a single string.
+    #[deprecated(
+        note = "please use Task::get_user_defined_attributes, legacy UDAs will be first class citizens in the future"
```
djmitche:
```suggestion
        note = "please use Task::get_user_defined_attributes"
```
I don't think we need to offer a reason here (and similarly for other places this appears)

docs/src/tasks.md
djmitche: I suspect the entire "Namespaced UDAs" section could be omitted. I don't know of anything that was using these, so I doubt there's much need for rationalization. Some of that text could perhaps be moved to the commit message and/or PR description.

grasegger:
@ryneeverett @djmitche thank you so much for your feedback. I removed the docs about namespaced UDAs and included the other comments.

ryneeverett (approved):

djmitche (approved):

djmitche:
Thanks!

PR #557, opened 2025-02-17
  Modify the working set in-place rather than clearing and rewriting

### LINKED ISSUE #414: Replace working set more efficiently
When renumbering tasks, we delete and re-add the entire working set:

https://github.com/GothenburgBitFactory/taskchampion/blob/e725a38c991dd32fd5f3b9b2daa4644b7dbbc9d2/taskchampion/src/taskdb/working_set.rs#L38-L42

This is pretty slow, and in most cases only a few tasks will change numbers. Optimize for that case, by only updating the items that have changed. This should allow removing the `Storage::clear_working_set` method as well.

djmitche:
This saves a lot of time in the common case of no change to the working set, as no changes are made. But the next-most-common cases are adding a new item (just one call to `add_to_working_set`) or removing a no-longer-pending item (a few calls to `set_working_set_item` for items  that have shifted their position).

This should be faster without any functional change.

Note that this includes #556, as without it cross_sync doesn't pass!

### DIFF
--- a/src/taskdb/working_set.rs
+++ b/src/taskdb/working_set.rs
@@ -10,6 +10,7 @@ pub(crate) fn rebuild<F>(txn: &mut dyn StorageTxn, in_working_set: F, renumber: 
 where
     F: Fn(&TaskMap) -> bool,
 {
+    let old_ws = txn.get_working_set()?;
     let mut new_ws = vec![None]; // index 0 is always None
     let mut seen = HashSet::new();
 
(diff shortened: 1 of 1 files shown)


src/taskdb/working_set.rs
```diff
@@ -152,4 +170,113 @@ mod test {
...
+        ops.push(Operation::Update {
+            uuid: uuids[0],
+            property: String::from("status"),
+            value: Some("pending".into()),
+            old_value: None,
+            timestamp: Utc::now(),
+        });
+        db.commit_operations(ops, |_| false)?;
+
+        // set the existing working_set as we want it, containing UUIDs 0 and 1.
```
ryneeverett: On this one we're adding all three UUID's to the working set.

ryneeverett (approved):

src/taskdb/working_set.rs
djmitche: Oh, I missed that comment - thanks!

PR #559, opened 2025-02-20
Update storage docs to match storage of sync'd operations

djmitche:
This was missed when the new task-history changes landed.

### DIFF
--- a/docs/src/storage.md
+++ b/docs/src/storage.md
@@ -24,7 +24,7 @@ In other words, operations act as deltas between database states.
 Operations are crucial to synchronization of replicas, described in [Synchronization Model](./sync-model.md).
 
 Operations are entirely managed by the replica, and some combinations of operations are described as "invalid" here.
-A replica must not create invalid operations, but should be resilient to receiving invalid operations during a synchronization operation.
+A replica must not create invalid operations, but should be resilient to receiving invalid operations during the synchronization process.
 
 Each operation has one of the forms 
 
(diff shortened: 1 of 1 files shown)


docs/src/storage.md
```diff
@@ -78,6 +78,11 @@ Replica operations are converted to sync operations as follows:
  * `Update(uuid, property, oldValue, newValue, timestamp)` -> `Update(uuid, property, newValue, timestamp)`
  * `UndoPoint()` -> Ø (dropped from operation sequence)
 
-Once a sequence of operations has been synchronized, there is no need to store those operations on the replica.
-The current implementation deletes operations at that time.
-An alternative approach is to keep operations for existing tasks, and provide access to those operations as a "history" of modifications to the task.
+Once a sequence of operations has been synchronized, they are not considered in subsequent sync operations.
```
ryneeverett: - In the context of the diff I understand why you changed this sentence this way but it's unclear to me what purpose it serves at the moment. Central to this is the ambiguous term "considered". Is this sentence trying to communicate that sync'ed operations are immutable? Maybe we could say that more explicitly?
- Can we come up with a less overloaded term to describe synchronization than "sync operations" -- maybe "sync process"? I noticed line 27 is even more guilty -- it uses the word operation three times: twice to describe taskchampion operations and once to describe the sync process.

docs/src/storage.md
```diff
@@ -78,6 +78,11 @@ Replica operations are converted to sync operations as follows:
...
+
+### Storage
+
+The storage backend stores all operations that apply to existing tasks, tracking which have and have not been synchronized.
+Only un-synchronized operations are considered when performing a sync operation.
```
ryneeverett: - This sentence is substantively identical to the last one in the previous paragraph so the same concerns apply. Not sure both are needed.
- "Unsynchronized" is a word which doesn't need to be hyphenated.

ryneeverett (approved):

docs/src/storage.md
djmitche: I think I can just delete this sentence, and keep the one in the "Storage" section.

PR #582, opened 2025-05-01
Make storage backends an optional (default enabled) feature

taliaferro:
Taskchampion's storage implementation is meant to be pluggable, providing the `Storage` and `StorageTxn` traits that a consumer of the library could implement to provide their own storage backend.

Someone in that scenario might want to disable the default SQLite storage backend implementation if they're not using it -- especially since it adds an external C dependency (libsqlite) which must be either bundled into the application or else already present on the target system.

This PR adds the new feature flags `storage` and `storage-sqlite` , which are enabled by default. Now if the library is built without default features, those default implementations will be absent.

### DIFF
--- a/src/server/config.rs
+++ b/src/server/config.rs
@@ -8,7 +8,7 @@ use crate::server::cloud::aws::AwsService;
 use crate::server::cloud::gcp::GcpService;
 #[cfg(feature = "cloud")]
 use crate::server::cloud::CloudServer;
-#[cfg(feature = "storage-sqlite")]
+#[cfg(feature = "server-local")]
 use crate::server::local::LocalServer;
 #[cfg(feature = "server-sync")]
 use crate::server::sync::SyncServer;
@@ -23,7 +23,7 @@ use uuid::Uuid;
 #[non_exhaustive]
 pub enum ServerConfig {
     /// A local task database, for situations with a single replica.
-    #[cfg(feature = "storage-sqlite")]
+    #[cfg(feature = "server-local")]
     Local {
         /// Path containing the server's DB
         server_dir: PathBuf,
@@ -97,7 +97,7 @@ impl ServerConfig {
     /// Get a server based on this configuration
     pub fn into_server(self) -> Result<Box<dyn Server>> {
         Ok(match self {
-            #[cfg(feature = "storage-sqlite")]
+            #[cfg(feature = "server-local")]
             ServerConfig::Local { server_dir } => Box::new(LocalServer::new(server_dir)?),
             #[cfg(feature = "server-sync")]
             ServerConfig::Remote {
--- a/src/server/config.rs
+++ b/src/server/config.rs
@@ -8,6 +8,7 @@ use crate::server::cloud::aws::AwsService;
 use crate::server::cloud::gcp::GcpService;
 #[cfg(feature = "cloud")]
 use crate::server::cloud::CloudServer;
+#[cfg(feature = "storage-sqlite")]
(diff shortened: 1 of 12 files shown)


taliaferro:
I've got the test suite passing with no default features; the only tests that require the sqlite backend are the server/sync ones so I put those tests behind the same feature flag.
Many of the tests use the in-memory backend, and it doesn't introduce new dependencies like sqlite, so I simply removed that feature flag and put the memory backend back in the no-default-features build.

I would appreciate some guidance on the doctests though; those are the only ones that are still failing. What's the best way to indicate that the code in a code block requires one of the default features?

src/server/config.rs
```diff
@@ -8,6 +8,7 @@ use crate::server::cloud::aws::AwsService;
 use crate::server::cloud::gcp::GcpService;
 #[cfg(feature = "cloud")]
 use crate::server::cloud::CloudServer;
+#[cfg(feature = "storage-sqlite")]
```
djmitche: This should probably be its own feature, since this isn't technically storage.

djmitche (commented):
This is a great idea -- thanks! Are you thinking of working on a different backend?

Something like this should do the trick for the doctests:

```diff
diff --git src/replica.rs src/replica.rs
index 4bb00ba5d..cfb7b085f 100644
--- src/replica.rs
+++ src/replica.rs
@@ -27,16 +27,18 @@ use uuid::Uuid;
 /// ## Operations
 ///
 /// Changes to a replica occur by committing [`Operations`]s. All methods that change a replica
 /// take an argument of type `&mut Operations`, and the necessary operations are added to that
 /// sequence. Those changes may be reflected locally, such as in a [`Task`] or [`TaskData`] value, but
 /// are not reflected in the Replica's storage until committed with [`Replica::commit_operations`].
 /**
 ```rust
+# #[cfg(feature = "storage-sqlite")]
+# {
 # use taskchampion::chrono::{TimeZone, Utc};
 # use taskchampion::{storage::AccessMode, Operations, Replica, Status, StorageConfig, Uuid};
 # use tempfile::TempDir;
 # fn main() -> anyhow::Result<()> {
 # let tmp_dir = TempDir::new()?;
 # let mut replica = Replica::new(StorageConfig::OnDisk {
 #   taskdb_dir: tmp_dir.path().to_path_buf(),
 #   create_if_missing: true,
@@ -50,16 +52,17 @@ t.set_description("my first task".into(), &mut ops)?;
 t.set_status(Status::Pending, &mut ops)?;
 t.set_entry(Some(Utc::now()), &mut ops)?;
 
 // Commit those operations to storage.
 replica.commit_operations(ops)?;
 #
 # Ok(())
 # }
+# }
 ```
 **/
 /// Undo is supported by producing an [`Operations`] value representing the operations to be
 /// undone. These are then committed with [`Replica::commit_reversed_operations`].
 ///
 /// ## Working Set
 ///
 /// A replica maintains a "working set" of tasks that are of current concern to the user,
```

djmitche (approved):
Looks good -- thanks!

taliaferro:
I am! I've been considering making a Web interface for my Taskwarrior database, using the browser's IndexedDB system as a local replica that can be used offline and synced with the normal sync server periodically. (That's why I'm trying to build Taskchampion without rusqlite; it's quite challenging to get it working in a browser WASM context.)

taliaferro:
And fantastic; thank you! I'll mark the PR as ready.

djmitche:
Very cool -- let me know if I can help further!

PR #583, opened 2025-05-03
Fix handling of URLs without trailing `/`

djmitche:
`Url::join` is sensitive to a trailing slash on the URL, and generates unexpected API URLs if this is not given. See
https://github.com/GothenburgBitFactory/taskwarrior/issues/3862.


djmitche:
This will fix https://github.com/GothenburgBitFactory/taskwarrior/issues/3862.

PR #589, opened 2025-06-09
Initial support for external hook handling

### LINKED ISSUE #3851: Force hooks to specify their target version
## Problem
Currently, hooks have the possibility to validate that the current hook version matches
their target (e.g., something like: `[ "$1" = "api:2" ] || { echo "Invalid version: $1"; exit 1; }`).

But this is completely voluntary, and not inspectable from the outside. Thus, TW has no
way to know which version a certain hook is targeting, and if it should warn the user
about outdated hooks.

## Proposed improvement
If we moved the hook naming scheme from `<type><hook_name>` to
`<type>-<version><hook_name>`, we would require hooks to specify what version they are
targeting.

The changes from a v2 hook directory to a v3:

### V2
```
$ tree ~/.config/task/hooks
/nix/store/qicdpkw69wzp8gxi52i73h4ddaqy5k14-mk-taskwarrior-hooks
├── on-add_enforce-projects
├── on-add_sync-git-repos
├── on-modify_sync-git-repos-mod
├── on-modify_track-timewarrior
└── on-modify_track-total-active-time

1 directory, 5 files
```

### V3
```
$ tree ~/.config/task/hooks
/nix/store/qicdpkw69wzp8gxi52i73h4ddaqy5k14-mk-taskwarrior-hooks
├── on-add-v3_enforce-projects
├── on-add-v3_sync-git-repos
├── on-modify-v3_sync-git-repos-mod
├── on-modify-v3_track-timewarrior
└── on-modify-v3_track-total-active-time

1 directory, 5 files
```

This could also be used to have a hook API version per type, but that seems to be
irrelevant for now.

bpeetz:
Fixes: https://github.com/GothenburgBitFactory/taskwarrior/issues/3851 https://github.com/GothenburgBitFactory/taskwarrior/issues/3838

Currently, following things are still left to be worked out (sorted in
their order of importance):
- [ ] The structured args, mentioned in https://github.com/GothenburgBitFactory/taskwarrior/issues/3840

- [x] The hook execute functions, print the hook feedback directly. They
  should obviously return in instead.

-  [ ] How to deal with the config access for hooks? (Ref: https://github.com/GothenburgBitFactory/taskwarrior/issues/3839)

- [x] The compose JSON function in `taskchampion` cannot convert UDAs based
  on their type (like TW did), because these types are stored in the TW
  config and not in the replica.
-> Should we find another way to convert them, or give the hooks
  exactly the UDA that is stored in the task?

- [ ] Tests (for compatibility with the previous TW task JSON composition)
  and a migration path are necessary.

- [x] Convenient function, that determines which hooks to run for a given `Operations` 

- [ ] Using the metadata and a `on-sync` hook.

- [x] The error handling used in `tw_hooks` and in the `from_composed_json`
  function uses gitoxide like small errors, instead of the shared error,
  previously used in taskchampion.


djmitche:
/cc @smemsh 

This is an interesting start! I appreciate the use of marker types for the hook kinds -- that's a nice design!

>  -> Should we find another way to convert them, or give the hooks exactly the UDA that is stored in the task?

This use of typed values has caused a lot of weird bugs, especially since the type can be changed independently of the value. For example if you define a string UDA, add it to some tasks, and then change the type to integer, things go very badly. So, I'd rather not replicate that problem in TaskChampion! It should be fine to leave interpretation of UDA values -- including handling of unparseable values -- to the hooks.

I need to re-read the linked issues to familiarize myself with the background, but I think it would help to have a description of how this works within the world of TaskChampion. Perhaps the most productive way to do that would be to edit the [book](https://gothenburgbitfactory.org/taskchampion/) content in `docs/`, adding a new chapter on hooks and discussing how they are run and how they are intended to be used.

bpeetz:
I've translate the v2 spec to the new format, and provided an access point for connecting hook execution to a replica commit. But I'm still unsure, what to do about the config access, as that is obviously impossible, but was quite convenient.

bpeetz:
And is there already a way to write end-to-end tests, that depend on file system access?
All the other tests, seem to just use a `InMemory` replica, but testing the hooks requires a file system structure.
Currently, I have a small test, pushed in the last commit, as binary, but that is obviously not a merge-able.

docs/src/hooks/spec.md
```diff
@@ -0,0 +1,216 @@
...
+
+When there are multiple scripts associated with the same event, the order of execution is the collating sequence of the script name.
+A hook script can be disabled by either removing execute permission, renaming it, or moving it to another location.
+
+The hook base path is client defined and should be taken from their respective
+documentation (e.g., Taskwarrior uses `rc.hooks.location`).
+
+## Input
+
+Input to hook scripts consists of JSON-formatted objects, one per line, in [JSON format](https://www.json.org/json-en.html).
```
djmitche, line 69: In light of GothenburgBitFactory/taskwarrior#3900, could we do something that would be less susceptible to corruption by shells? Maybe base64-encoded JSON?

docs/src/hooks/spec.md
```diff
@@ -0,0 +1,216 @@
...
+
+This contained keys are dependent on the event (e.g., `on-commit` provides JSON encoded
+operations, whilst `on-add` supplies a task encoded as JSON).
+
+A JSON encoded task might look like:
+
+```
+{"description":"Buy some milk","entry":"20141118T050231Z","status":"pending","uuid":"a360fc44-315c-4366-b70c-ea7e7520b749"}
```
djmitche, line 77: That entry is in the wrong format -- entry is a [number of seconds since the epoch](https://gothenburgbitfactory.org/taskchampion/tasks.html#keys). Also, the `uuid` wouldn't typically be included here. If it's necessary to send that to the hook, perhaps something like 
```json
{"task": {"description":"Buy some milk","entry":"1693329505","status":"pending"},"uuid":"a360fc44-315c-4366-b70c-ea7e7520b749"}
```
this would allow adding additional keys a the top level, and would also mean that every hook could accept a single JSON object as input, rather than JSON-lines. For example, `on-commit` could be `{"operations": [..]}`.

docs/src/hooks/spec.md
```diff
@@ -0,0 +1,216 @@
...
+
+The hook base path is client defined and should be taken from their respective
+documentation (e.g., Taskwarrior uses `rc.hooks.location`).
+
+## Input
+
+Input to hook scripts consists of JSON-formatted objects, one per line, in [JSON format](https://www.json.org/json-en.html).
+
+This contained keys are dependent on the event (e.g., `on-commit` provides JSON encoded
+operations, whilst `on-add` supplies a task encoded as JSON).
```
djmitche, line 72: Let's link to the task-model page here -- we want hooks authors to be aware that the keys and values are as specified in that model, not TW's JSON format.

docs/src/hooks/spec.md
```diff
@@ -0,0 +1,216 @@
...
+| `on-launch` | This event is triggered at instantiation of the replica, or at call time of `with_hook_base_path`. This event provides an opportunity to prevent further processing.                                                                      |
+| `on-exit`   | This event is triggered at drop time of the replica. This event has no opportunity to modify tasks.                                                                                                                                       |
+| `on-add`    | This event is triggered when a new task is created, just before the task is committed. This event provides an opportunity to approve, deny, or modify the task being created.                                                             |
+| `on-modify` | This event is triggered when an existing task is modified, just before the task is committed. This event provides an opportunity to approve, deny, or further change the modification of the task.                                        |
```
djmitche, line 31: Are these called once for each modified task when `on-commit` is invoked, based on the set of operations?
 - If the ops for a given task include `Create`, invoke `on-add` with the state of the task after those operations are applied.
 - If the ops for a given task do not include `Create`, invoke `on-modify` instead.

docs/src/hooks/spec.md
```diff
@@ -0,0 +1,216 @@
...
+| `on-commit` | This event is triggered when operations are committed. Instead of `on-add` or `on-modify` the task has not be (crudely) constructed from operations.
```
djmitche, line 32: "crudely" makes me nervous. I think we should have a precise definition of what happens, that is future compatible. A "crude" assembly is the sort of thing that will work in tests and for simple hooks, and then fail for more complex cases that arise long after release.

djmitche: Also, is this called before or after the operations are committed?

docs/src/hooks/spec.md
```diff
@@ -0,0 +1,216 @@
...
+{"description":"Buy some milk","entry":"20141118T050231Z","status":"pending","uuid":"a360fc44-315c-4366-b70c-ea7e7520b749"}
+```
+
+Before the event-specific input the hook script will be supplied with an object of metadata.
+
+The following fields are present:
+
+| Field name                     | Meaning                                                                                         | Example value                                       |
+|--------------------------------|-------------------------------------------------------------------------------------------------|-----------------------------------------------------|
+| `db_location`                  | Path to the database location with which this replica has been initialized.                     | `/home/user/.local/share/task/taskchampion.sqlite3` |
```
djmitche, line 86: Two issues here:
 - I suspect the DB will be locked when this call is made.
 - SQLite is only one storage backend, so this will need to be able to handle other backends as well (the only alternative is InMemory right now, but I expect we'll have MySQL, Postgres, etc. backends eventually)

docs/src/hooks/spec.md
```diff
@@ -0,0 +1,216 @@
...
+
+## Output
+
+Output from a hook script is a combination of a JSON and text.
+Whitespace within the line is permitted, but the JSON or text output *must* occupy only one line of output.
+As it is read back line-wise.
+
+Feedback text is optional for a successful hook script, but required for a failing script.
+If provided it will be logged by Taskchampion (error level on failure, or info on
+success).
```
djmitche, line 101: This makes the feedback difficult to capture for an application embedding TaskChampion (e.g., a webserver) -- it would have to use a mutex to serialize _all_ TC calls that might trigger a hook, and add a log interceptor, and then correlate any log messages to the running call.

docs/src/hooks/spec.md
```diff
@@ -0,0 +1,216 @@
...
+| `on-add`    | This event is triggered when a new task is created, just before the task is committed. This event provides an opportunity to approve, deny, or modify the task being created.                                                             |
+| `on-modify` | This event is triggered when an existing task is modified, just before the task is committed. This event provides an opportunity to approve, deny, or further change the modification of the task.                                        |
+| `on-commit` | This event is triggered when operations are committed. Instead of `on-add` or `on-modify` the task has not be (crudely) constructed from operations.
+                Use this if you need fine-grained control over the committed operations.  This event provides an opportunity to approve, deny, or further change the operations.                                                                          |
+
+Task completion and deletion are considered modifications, indicated by the change to the 'status' attribute.
+
+
+## File structure
```
djmitche, line 38: Hosted solutions (e.g., a webserver handling lots of users' tasks) would need some other option than direct filesystem access here. I think we'll need an abstraction which those services can use, allowing for example to store scripts in a database. I'm not sure how those will _execute_ those scripts -- maybe WASM?

docs/src/hooks/spec.md
```diff
@@ -0,0 +1,216 @@
...
+
+Currently, that happens according to the following algorithm:
+
+1. All operations are sorted into separate stacks according to their associated UUID.
+1. Every stack's operations are sorted (`Created`, then `Update`)
+1. The operations are folded into a single task object (i.e., updates are applied).
+1. If the stack started with a `Create` operation, the task is categorized as “added”
+   otherwise it is categorized as “modified”.
+
+The returned tasks are turned back into operations with the following algorithm.
```
djmitche, line 205: Maybe this would be better handled by the hook producing a set of operations to apply?

djmitche (commented):
Thanks for the docs!

It seems like some of the awkward bits highlighted in my comments below are things that TaskWarrior already handles in how it wraps TaskChampion (e.g., translating stored epoch timestamps into ISO timestamps or defining a "modification"). That's some pretty serious abstraction boundary leakage! My initial reaction is, TaskWarrior should continue to handle those wrappings. But I think the whole point of this exercise, based on discussion inhttps://github.com/GothenburgBitFactory/taskwarrior/issues/3838, is to take TaskWarrior out of the picture so that a TaskChampion replica embedded in another application, for example, Bugwarrior, an Android app, a hand-written Python script using taskchampion-py, or a web service, could still execute the same hooks. I don't see clearly how this proposal would work in all of those situations.

I'm sorry to be so high-level and vague about this, but I think this PR and associated issue is still missing a clear statement of the problem it is solving. It introduces a lot of costs (filesystem access, incompatibility with other storage backends, high potential for bugs and incompatibilities) and it's difficult to see how to balance those against unclear benefits.

My best understanding of the problem is, it's currently possible for a script using something like `taskchampion-py` to modify a replica that is also accessed via TaskWarrior, and modifications by that script do not trigger hooks. To me, it's not clear this proposal fixes that issue, and the fix seems not worth the cost.

Also, this doesn't seem to address https://github.com/GothenburgBitFactory/taskwarrior/issues/3851, although it could address that issue with minor changes.

Anyway, I'd be happy to talk about this in a more interactive setting, if that's helpful.

docs/src/hooks/spec.md
bpeetz: Oh, I just copied that example from the v1 docs (on the TW website) and didn't notice these differences.

The actual implementation, in its current state, does not actually pass the UUID and provides timestamps as seconds since the Epoch.

I will update the docs, so that they are in sync with that.

The single-object per hook seems like a good idea overall.

docs/src/hooks/spec.md
bpeetz: Would this be fixed by just passing one json object?

This should allow passing newlines (although the common `read -r json` pattern, used in the various shell example hooks, would no longer work and would need replacing with something like `jq`. This seems still like a better way though.)

docs/src/hooks/spec.md
bpeetz: > SQLite is only one storage backend, so this will need to be able to handle other backends as well

What about mirroring the `Storage` API?

That would mean adding a `Hooks` trait and renaming the current impl to something like `TwHooks`. This would also allow a nice entry point for an `InMemoryHooks` struct, which could be used for tests (and I guess for a hosted use-case).

> I suspect the DB will be locked when this call is made.

A better solution would probably be to not include the path at all and provide a way for the library users (e.g. TW) to add extra meta-data (like its configuration as json object or the config file location and such things). Maybe simply via environment variables, or as a json object, that is included in the one object sent to the hook?

docs/src/hooks/spec.md
bpeetz: Returning an `Output` struct (similar to `std`'s `process::Output`), with separate `feedback` and `json` fields would probably be more adaptable?

I initially decided log the feedback directly, because I expected most library users to do this anyways.

docs/src/hooks/spec.md
bpeetz: I think that would be covered by providing a way for users to bring their own Hook adapter (i.e., the `Hooks` trait I mentioned [here](https://github.com/GothenburgBitFactory/taskchampion/pull/589#discussion_r2190448381))?

docs/src/hooks/spec.md
djmitche: Returning the output, or using a logging callback, would be good solutions.

docs/src/hooks/spec.md
djmitche: Yeah, abstracting it a bit makes sense, and might help give a clearer vision of what the abstraction is.

docs/src/hooks/spec.md
djmitche: > A better solution would probably be to not include the path at all and provide a way for the library users (e.g. TW) to add extra meta-data (like its configuration as json object or the config file location and such things). Maybe simply via environment variables, or as a json object, that is included in the one object sent to the hook?

I like this idea, but that makes the hooks specific to the library user again, at which point why not just call the hooks from the user? For example, an app that listens for GitHub webhooks and updates tasks corresponding to GitHub issues wouldn't be able to provide the same context as TaskWarrior, so even if it was running the same hook scripts as TaskWarrior, they would behave differently.

bpeetz:
I am unfortunately unable to give this feature the time that it deserves and to push it through (especially because my time will become more sparse in the foreseeable future).

Therefore, as I don't want to block anyone, I'll close this pr.
If anyone wants to pick this up, feel free to use this pr as a starting point.
I will definitely continue to watch this space and might pick up work on this again in the future.

djmitche:
Thank you, @bpeetz, for your work on this! It is, indeed, an expansive topic and deserves a good bit of time.

PR #593, opened 2025-07-06
Buffer input to Zlib to improve perf

### LINKED ISSUE #591: Snapshots are large and take a long time to generate
Several times now I've wondered "why is task sync taking so long" and as it turns out, it's because it's generating a snapshot. Let's make these faster to generate and/or smaller.

djmitche:
The JSON encoding process necessarily makes a lot of small writes. It seems that ZLib handles these poorly -- it is very slow and uses gobs of memory. Wrapping this in BufWriter buffers the input into 8KiB chunks, dramatically improving performance.

Fixes #591.

### DIFF
--- a/src/taskdb/snapshot.rs
+++ b/src/taskdb/snapshot.rs
@@ -57,8 +57,14 @@ impl<'de> Deserialize<'de> for SnapshotTasks {
 
 impl SnapshotTasks {
     pub(super) fn encode(&self) -> Result<Vec<u8>> {
-        let mut encoder = ZlibEncoder::new(Vec::new(), Compression::default());
+        let encoder = ZlibEncoder::new(Vec::new(), Compression::default());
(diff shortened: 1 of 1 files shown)


ryneeverett (commented):
Per [the docs](https://doc.rust-lang.org/std/io/struct.BufWriter.html), don't we need to manually flush the BufWriter?

> It is critical to call flush before BufWriter<W> is dropped. Though dropping will attempt to flush the contents of the buffer, any errors that happen in the process of dropping will be ignored. Calling flush ensures that the buffer is empty and thus dropping will not even attempt file operations.

ryneeverett (approved):
Oh, I see now that the comment in the example for `into_inner` implies that it flushes the buffer. Not sure I'm impressed with the Rust docs here but seems like you got it right.

djmitche:
I saw the same docs and they were clear enough for me, but I see your point. Thanks for looking deeply!

Issue #599, opened 2025-07-30
Add WASM/async support via feature flag


geofflittle:
Hello! I'm working on a project that compiles to WASM and would love to use `taskchampion` as a library. Currently, this isn't possible due to a dependency that doesn't support the `wasm32-unknown-unknown` target.

This issue proposes a non-breaking change to add first-class support for using `taskchampion` in a WASM and `async` context.

**The Problem**
When attempting to compile `taskchampion` for a WASM target, the build fails:

* **UUID Generation:** The `uuid` crate's default `v4` feature relies on OS-level random number generation, which is not available in WASM.

**Proposed Solution**
I propose introducing a new `wasm` feature flag. When a user enables this feature and disables default features, the library will be fully WASM-compatible.

A consumer of the library would use it in their Cargo.toml like this:

```toml
[dependencies]
taskchampion = { version = "...", default-features = false, features = ["wasm"] }
```

This approach is fully backward-compatible and requires no changes for existing users.

**Implementation Plan**
I plan to submit a pull request with several small, logical commits to implement this. Each commit will represent a focused step in the process, making the changes easier to review.

1. **Add `wasm` feature and configure `uuid`**
    * Add the new `wasm = ["uuid/js"]` feature to `Cargo.toml`.
    * This configures the `uuid` crate to use the browser's random number generator when the `wasm` feature is active.
2. **Introduce the `AsyncServer` trait**
    * Add the `async-trait` dependency.
    * Define a new `pub trait AsyncServer`, mirroring the synchronous `Server` trait, to provide a clear interface for non-blocking server implementations.
3. **Implement `async` synchronization**
    * Implement the core async logic in a new `taskdb/async_sync.rs` module.
    * Add the public-facing `async fn sync_async(...)` methods to `Replica` and `TaskDb`.

This will make taskchampion a much more versatile library and open it up to a new ecosystem of web and serverless applications. I'm happy to discuss this proposal and am open to any feedback.

Thank you!
-Geoff

geofflittle:
After further integration testing with my local changes, I've found that the library would be more robust and easier to use if it included a concrete `AsyncSyncServer` implementation directly, rather than requiring downstream async users to build their own from scratch.

This keeps the library's encryption logic properly encapsulated and provides a valuable, reusable component for the community.

My updated plan is to add the following commit:

4. **Add AsyncSyncServer Implementation**
    * Add `reqwest` as an optional dependency.
[...]

PR #600, opened 2025-08-01
Refactor for upcoming async support

geofflittle:
Related to #599. This is a pre-requisite to support async.

This change moves transaction control up from TaskDb to Replica. This makes the TaskDb layer stateless by having its methods accept a transaction as an argument, while Replica becomes responsible for initiating the transactions.

This decoupling is needed for the next step, which will be to introduce a callable pattern at the Replica layer to manage transaction lifetimes. That will lead to the second-next step, introducing async.

### DIFF
--- a/src/replica.rs
+++ b/src/replica.rs
@@ -2,7 +2,7 @@ use crate::depmap::DependencyMap;
 use crate::errors::Result;
 use crate::operation::{Operation, Operations};
 use crate::server::Server;
-use crate::storage::{Storage, TaskMap};
+use crate::storage::{Storage, StorageTxn, TaskMap};
 use crate::task::{Status, Task};
 use crate::taskdb::TaskDb;
 use crate::workingset::WorkingSet;
@@ -70,6 +70,8 @@ replica.commit_operations(ops)?;
 /// pending tasks are automatically added to the working set, and the working set can be
 /// "renumbered" when necessary.
 pub struct Replica {
+    storage: Box<dyn Storage>,
+
     taskdb: TaskDb,
 
     /// If true, this replica has already added an undo point.
@@ -82,7 +84,8 @@ pub struct Replica {
 impl Replica {
     pub fn new(storage: Box<dyn Storage>) -> Replica {
         Replica {
-            taskdb: TaskDb::new(storage),
+            storage,
+            taskdb: TaskDb::new(),
             added_undo_point: false,
             depmap: None,
         }
@@ -117,17 +120,19 @@ impl Replica {
         };
         task.update(property, value, &mut ops);
         self.commit_operations(ops)?;
+        let mut txn = self.storage.txn()?;
         Ok(self
             .taskdb
-            .get_task(uuid)?
+            .get_task(txn.as_mut(), uuid)?
             .expect("task should exist after an update"))
     }
 
     /// Get all tasks represented as a map keyed by UUID
     pub fn all_tasks(&mut self) -> Result<HashMap<Uuid, Task>> {
         let depmap = self.dependency_map(false)?;
+        let mut txn = self.storage.txn()?;
         let mut res = HashMap::new();
-        for (uuid, tm) in self.taskdb.all_tasks()?.drain(..) {
+        for (uuid, tm) in self.taskdb.all_tasks(txn.as_mut())?.drain(..) {
             res.insert(uuid, Task::new(TaskData::new(uuid, tm), depmap.clone()));
         }
         Ok(res)
@@ -135,8 +140,9 @@ impl Replica {
 
     /// Get all task represented as a map of [`TaskData`] keyed by UUID
     pub fn all_task_data(&mut self) -> Result<HashMap<Uuid, TaskData>> {
+        let mut txn = self.storage.txn()?;
         let mut res = HashMap::new();
-        for (uuid, tm) in self.taskdb.all_tasks()?.drain(..) {
+        for (uuid, tm) in self.taskdb.all_tasks(txn.as_mut())?.drain(..) {
             res.insert(uuid, TaskData::new(uuid, tm));
         }
         Ok(res)
@@ -144,7 +150,8 @@ impl Replica {
 
     /// Get the UUIDs of all tasks
     pub fn all_task_uuids(&mut self) -> Result<Vec<Uuid>> {
-        self.taskdb.all_task_uuids()
+        let mut txn = self.storage.txn()?;
+        self.taskdb.all_task_uuids(txn.as_mut())
     }
 
     /// Get an array containing all pending tasks
@@ -160,9 +167,10 @@ impl Replica {
     }
 
     pub fn pending_task_data(&mut self) -> Result<Vec<TaskData>> {
+        let mut txn = self.storage.txn()?;
         let res = self
             .taskdb
-            .get_pending_tasks()?
+            .get_pending_tasks(txn.as_mut())?
             .into_iter()
             .map(|(uuid, taskmap)| TaskData::new(uuid, taskmap))
             .collect::<Vec<_>>();
@@ -173,7 +181,12 @@ impl Replica {
     /// Get the "working set" for this replica.  This is a snapshot of the current state,
     /// and it is up to the caller to decide how long to store this value.
     pub fn working_set(&mut self) -> Result<WorkingSet> {
-        Ok(WorkingSet::new(self.taskdb.working_set()?))
+        let mut txn = self.storage.txn()?;
+        Replica::working_set_with_txn(&mut self.taskdb, txn.as_mut())
+    }
+
+    fn working_set_with_txn(taskdb: &mut TaskDb, txn: &mut dyn StorageTxn) -> Result<WorkingSet> {
+        Ok(WorkingSet::new(taskdb.working_set(txn)?))
     }
 
     /// Get the dependency map for all pending tasks.
@@ -191,6 +204,7 @@ impl Replica {
     /// Calculating this value requires a scan of the full working set and may not be performant.
     /// The [`TaskData`] API avoids generating this value.
     pub fn dependency_map(&mut self, force: bool) -> Result<Arc<DependencyMap>> {
+        let mut txn = self.storage.txn()?;
         if force || self.depmap.is_none() {
             // note: we can't use self.get_task here, as that depends on a
             // DependencyMap
@@ -198,13 +212,13 @@ impl Replica {
             let mut dm = DependencyMap::new();
             // temporary cache tracking whether tasks are considered Pending or not.
             let mut is_pending_cache: HashMap<Uuid, bool> = HashMap::new();
-            let ws = self.working_set()?;
+            let ws = Replica::working_set_with_txn(&mut self.taskdb, txn.as_mut())?;
             // for each task in the working set
             for i in 1..=ws.largest_index() {
                 // get the task UUID
                 if let Some(u) = ws.by_index(i) {
                     // get the task
-                    if let Some(taskmap) = self.taskdb.get_task(u)? {
+                    if let Some(taskmap) = self.taskdb.get_task(txn.as_mut(), u)? {
                         // search the task's keys
                         for p in taskmap.keys() {
                             // for one matching `dep_..`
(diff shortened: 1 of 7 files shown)


src/errors.rs
```diff
@@ -36,11 +36,21 @@ macro_rules! other_error {
...
 other_error!(crate::storage::sqlite::SqliteError);
+#[cfg(feature = "storage-sqlite")]
+other_error!(tokio_rusqlite::Error);
+
+#[cfg(feature = "storage-sqlite")]
+impl From<Error> for tokio_rusqlite::Error {
+    fn from(e: Error) -> Self {
+        tokio_rusqlite::Error::Other(e.into())
+    }
+}
```
djmitche: This is something we can look at later, but this probably shouldn't be necessary! I think the main reason this is required is the use of `con.call(..)`, which expects a closure returning a result with `tokio_rusqlite::Error`. I expect in most cases, the errors are already `tokio_rusqlite::Error`, then converted to `crate::Error`, then using this `From` implementation to convert it back. So, there's probably some optimization to do!

If this _is_ required in places, then we will probably want to do a lossless conversion back, rather than the wrapping approach here, which might end up with a `tokio_rusqlite::Error::Other(crate::Error::Other(tokio_rusqlite::Error::SomethingElse))`.

src/storage/sqlite/schema.rs
```diff
@@ -26,8 +27,8 @@ type UpgradeFn = fn(&Transaction) -> Result<()>;
 const VERSIONS: &[(DbVersion, UpgradeFn)] = &[(DbVersion(0, 1), upgrade_to_0_1)];
 pub(super) const LATEST_VERSION: DbVersion = VERSIONS[VERSIONS.len() - 1].0;
 
-pub(super) fn upgrade_db(con: &mut Connection) -> Result<()> {
-    let mut current_version = get_db_version(con)?;
+pub(super) async fn upgrade_db(con: &mut Connection) -> Result<()> {
```
djmitche: I suspect this always returns a `Result<(), tokio_rusqlite::Error>`

src/storage/async_storage/adapter.rs
```diff
@@ -0,0 +1,145 @@
...
+    /// The actual async storage implementation, boxed to handle any type
+    /// that implements the `AsyncStorage` trait.
+    inner: Box<dyn AsyncStorage>,
+}
+
+impl BlockingStorageAdapter {
+    /// Create a new BlockingStorageAdapter, wrapping the given async storage backend.
+    ///
+    /// This will build a new single-threaded Tokio runtime to run the backend on.
+    pub(crate) fn new(storage: Box<dyn AsyncStorage>) -> Result<Self> {
```
djmitche: It'd be a bit more ergonomic to take the async storage by value (`storage: impl AsyncStorage`) and box it for the caller.

src/storage/async_storage/adapter.rs
```diff
@@ -0,0 +1,145 @@
+// src/storage/adapter.rs
```
djmitche: I actually like this filename a bit better -- I don't think the async functionality needs to be isolated in a submodule.

src/storage/async_storage/mod.rs
```diff
@@ -0,0 +1,92 @@
...
+use crate::server::VersionId;
+use crate::storage::TaskMap;
+use async_trait::async_trait;
+use uuid::Uuid;
+
+pub(crate) mod adapter;
+
+/// An async Storage service, allowing storage operations.
+#[async_trait]
+pub(crate) trait AsyncStorage: Send + Sync {
```
djmitche: Part of the reason to do this now is to make the public trait -- which users might implement -- async, as a breaking change along with the other changes discussed in #599 like making `Replica::sync` async. So,
 * This trait should be named `Storage` (and similarly `StorageTxn`), with the blocking trait renamed to `BlockingStorage`.
 * It should be pub.
 * The existing crate (now `BlockingStorage`) should be `pub(crate)` so it's not user-visible
 * Since it will be the _only_ such trait in the future (when support for the blocking trait is dropped), it should have pride of place in `src/storage/mod.rs` where the existing trait is.

The overall effect from an external (`pub`) perspective is that all of the methods in the `Storage` traits became async.

src/storage/sqlite.rs
```diff
@@ -77,18 +78,18 @@ impl ToSql for Operation {
     }
 }
 
-/// SqliteStorage is an on-disk storage backed by SQLite3.
-pub(super) struct SqliteStorage {
+/// AsyncSqliteStorage is an on-disk storage backed by SQLite3.
+pub(super) struct AsyncSqliteStorage {
```
djmitche: I don't think we need to rename this.

src/storage/async_storage/mod.rs
```diff
@@ -0,0 +1,92 @@
...
+    /// Get an (immutable) task, if it is in the storage
+    async fn get_task(&mut self, uuid: Uuid) -> Result<Option<TaskMap>>;
```
djmitche: ❗⚠️❗ Hang on.. what happened to the transaction??  

A key part of the storage functionality is that transactions are applied atomically. For example, if an error happens after a call to `set_task`  but before the corresponding `add_operation`, then the result of `set_task` must be rolled back and never visible in the DB. 

This proved a bit difficult with futures and lifetimes and Send and Sync when I made a similar refactor in https://github.com/GothenburgBitFactory/taskchampion-sync-server/pull/126 but I did get it to work.

src/storage/test.rs
```diff
@@ -1,717 +1,530 @@
...
+use crate::storage::async_storage::AsyncStorage;
 use crate::storage::{taskmap_with, DEFAULT_BASE_VERSION};
 use crate::Operation;
 use chrono::Utc;
 use pretty_assertions::assert_eq;
 use uuid::Uuid;
 
 /// Define a collection of storage tests that apply to all storage implementations.
-macro_rules! storage_tests {
+macro_rules! async_storage_tests {
```
djmitche: No need to rename this, since all storage is async now.

djmitche (commented):
This is looking good! Some of my comments are a matter of preference, but the transaction comment is a big deal.

src/errors.rs
geofflittle: I'm with you!

This commit uses this `From` impl to convert `anyhow::Error` from calls to `context()` to `tokio_rusqlite:Error` for the Connection closure, as you noted. Some `context` calls are in the sqlite files. Here's an example that uses the From<Error> code.

```rust
con.call(
    |c: &mut rusqlite::Connection| -> tokio_rusqlite::Result<()> {
        let res: rusqlite::Result<()> =
            c.query_row("PRAGMA journal_mode=WAL", [], |_| Ok(()));
        let context: anyhow::Result<()> = res.context("Setting journal_mode=WAL");
        context.map_err(Error::from)?; // The added From<Error> impl is being used here
        Ok(())
    },
)
.await?;
```

Let me know what's your preference for converting `anyhow::Error` to `tokio_rusqlite::Error`. If the `context` call should really be outside `con.call`, let me know how you'd like to handle instances where there are multiple calls to `context` within ` a `con.call`, like in `sync_complete`.

src/storage/sqlite/schema.rs
geofflittle: Forgive me--you're saying you want this to always return `tokio_rusqlite::Result<()>` / should be changed to always return this?

src/storage/async_storage/adapter.rs
geofflittle: Indeed!

src/errors.rs
djmitche: I think it's OK to omit the context calls when the expected return type is `tokio_rusqlite::Result`. They're a debugging convenience, but I think it's better to have the real error than a deeply nested version.

src/storage/sqlite/schema.rs
djmitche: Yes, and similarly I think a lot of functions that are private to `crate::storage::sqlite` can just use `tokio_rusqlite::Result` and should be changed to just do that (including removing the `context` calls).

djmitche:
There was some chat discussion of the transaction boundary, but I was on my phone so I apologize for being brief.

It's _absolutely critical_ that the DB be consistent:
 * An operation is in `operations` if and only if it has also been applied to `tasks`
 * The sync version identified by the `base_version` in `sync_metadata` has been applied to both `operations` and `tasks`. Any versions _after_ that `base_version` have not been applies.
 * Any operations marked as `synced` have been successfully sent to a server (the converse is not true -- it's possible to send a version to the server and crash before getting a response, which results in later downloading and "applying" that version -- the operational transforms sync process handles this situation)

Less critical things (but it'd be OK if these are guaranteed):
 * WorkingSet is updated after any status changes received via sync
 * Expiration of old tasks and concurrent modifications of those tasks in the same replica.

In general, I think it'd be fine to make transactions a pretty high-level concept and handle them async. However, we still need to be able to wrap methods in a sync call, for use in other languages. For example, we need to be able to call `Replica::commit_operations` or `Replica::all_task_data` from C++. It's OK to require a small wrapper in the C++ glue (e.g., `block_on(replica.all_task_data())`, but wrapping async callbacks into C++, Python, Ruby, etc. would be prohibitively difficult.

geofflittle:
Hi @djmitche,

I've changed this PR to contain the first step in the refactoring needed to prepare the library for the `async`/WASM implementation.

The core change is moving transaction control up from `TaskDb` to `Replica`. This makes the `TaskDb` layer stateless by having its methods accept a transaction as an argument, while `Replica` becomes responsible for initiating the transactions.

This decoupling is a prerequisite for the next step, which will be to introduce a callable pattern at the Replica layer to manage transaction lifetimes. That will lead to the second-next step, introducing `async`.

Looking forward to your feedback on this change before I proceed with the next steps! Once I get your feedback, I'll address your comments and then continue to add further commits to this PR.

src/taskdb/mod.rs
```diff
@@ -16,21 +16,19 @@ mod working_set;
 /// A TaskDb is the backend for a replica.  It manages the storage, operations, synchronization,
```
djmitche:
```suggestion
/// A TaskDb is the backend for a replica.  It manages the operations, synchronization,
```

src/taskdb/mod.rs
```diff
@@ -16,21 +16,19 @@ mod working_set;
...
 /// and so on, and all the invariants that come with it.  It leaves the meaning of particular task
 /// properties to the replica and task implementations.
-pub(crate) struct TaskDb {
-    storage: Box<dyn Storage>,
-}
+pub(crate) struct TaskDb;
```
djmitche, line 19: This type is kind of unnecessary now, but we can factor it away in a followup, just to limit the number of changes in this PR.

djmitche (commented):
This is looking great! It's probably worth renaming this PR and updating its description before committing -- and doing the followup work (to actually make it async) in another PR.

src/taskdb/mod.rs
geofflittle: Heard!

src/taskdb/mod.rs
```diff
@@ -13,24 +13,22 @@ mod sync;
...
-    #[cfg(test)]
-    pub(crate) fn new_inmemory() -> TaskDb {
-        use crate::StorageConfig;
-        TaskDb::new(StorageConfig::InMemory.into_storage().unwrap())
-    }
+    // #[cfg(test)]
+    // pub(crate) fn new_inmemory() -> TaskDb {
+    //     use crate::StorageConfig;
+    //     TaskDb::new(StorageConfig::InMemory.into_storage().unwrap())
+    // }
```
djmitche: Please remove!

src/replica.rs
```diff
@@ -399,10 +417,9 @@ impl Replica {
     /// system
     pub fn sync(&mut self, server: &mut Box<dyn Server>, avoid_snapshots: bool) -> Result<()> {
         self.taskdb
-            .sync(server, avoid_snapshots)
+            .sync(self.storage.txn()?.as_mut(), server, avoid_snapshots)
             .context("Failed to synchronize with server")?;
-        self.rebuild_working_set(false)
-            .context("Failed to rebuild working set after sync")?;
+        self.rebuild_working_set(false)?;
```
djmitche: This seems to have lost the context?

```suggestion
        self.rebuild_working_set(false)
            .context("Failed to rebuild working set after sync")?;
```

djmitche (changes requested):
Two fairly trivial changes, after which we can merge this!

djmitche (approved):
Awesome!

djmitche:
(oops, this comment was supposed to be on #599)

Issue #546, opened 2025-02-02
Sync to custom S3 apis


Juliaria08:
Similar to the Azure sync that would be implemented in https://github.com/GothenburgBitFactory/taskchampion/issues/369, the GCP sync implemented in https://github.com/GothenburgBitFactory/taskwarrior/issues/3185, and the AWS sync implemented in https://github.com/GothenburgBitFactory/taskchampion/issues/368, we should be able to sync replicas to a S3-compatible object storages like [MinIO](https://github.com/minio/minio), [seaweedfs](https://github.com/seaweedfs/seaweedfs/wiki/Amazon-S3-API), [Garage](https://garagehq.deuxfleurs.fr/), [Versity's S3 Gateway](https://github.com/versity/versitygw), or even more-enterprise solutions like [Tigris](https://www.tigrisdata.com/).

I'm not sure how easy it'd be to allow the use of a custom endpoint, nor if it's already implemented by using the `aws` cli client option, but I tried setting the `AWS_ENDPOINT_URL` environment-variable to `https://play.min.io:9000`, having the `aws` client configured properly, and it didn't work IIRC.

djmitche:
This is a good idea! If you can figure out how to do it with the normal AWS shell command (profile config??), it should be pretty easy to replicate that in the TW config options.

Juliaria08:
> This is a good idea! If you can figure out how to do it with the normal AWS shell command (profile config??), it should be pretty easy to replicate that in the TW config options.

The [docs](https://min.io/docs/minio/linux/integrations/aws-cli-with-minio.html) say to use `aws --endpoint-url http://server.url.here.tld.invalid s3 ls`

Suppposedly, there's also a way to do it by setting environment variables, or editing the config (manually?): <https://aws.amazon.com/blogs/developer/new-improved-flexibility-when-configuring-endpoint-urls-with-the-aws-sdks-and-tools/>

djmitche:
It looks like this can be included in a profile: https://docs.aws.amazon.com/cli/v1/userguide/cli-configure-files.html#cli-config-endpoint_url

If that works, then that might be the best option -- we can document how to set up the profile, but avoid adding purpose-specific config to TaskChampion and Taskwarrior.

If that doesn't work, could you check if a patch like this works?
```diff
diff --git src/server/cloud/aws.rs src/server/cloud/aws.rs
index 264b6824b..3c64460c3 100644
--- src/server/cloud/aws.rs
[...]

PR #601, opened 2025-08-03
Add support for s3-compatible storage

travisby:
(Will add description and move out of draft)

### DIFF
--- a/src/server/cloud/aws.rs
+++ b/src/server/cloud/aws.rs
@@ -101,7 +101,7 @@ impl AwsService {
                 // 3. If no region is discovered, hardcode to "us-east-1"
                 //
                 // If there's a region specified we will always prefer that
-                // next, the default provider chain will look at things like AWS_REGION environment
+                // Next, the default provider chain will look at things like AWS_REGION environment
                 // variables, the profile file, etc.
                 //
                 // we provide the hardcoded fallback because a region MUST be set
--- a/src/server/cloud/aws.rs
+++ b/src/server/cloud/aws.rs
@@ -13,7 +13,7 @@ use aws_sdk_s3::{
 use std::future::Future;
 use tokio::runtime::Runtime;
 
-/// A [`Service`] implementation based on the Google Cloud Storage service.
+/// A [`Service`] implementation based on the AWS Simple Storage Service.
 pub(in crate::server) struct AwsService {
     client: s3::Client,
     rt: Runtime,
@@ -63,9 +63,11 @@ pub enum AwsCredentials {
 
 impl AwsService {
     pub(in crate::server) fn new(
-        region: String,
+        region: Option<String>,
         bucket: String,
         creds: AwsCredentials,
+        endpoint_url: Option<String>,
+        force_path_style: bool,
     ) -> Result<Self> {
         let rt = Runtime::new()?;
 
@@ -92,13 +94,36 @@ impl AwsService {
                         // Just use the default.
                     }
                 }
-                config_provider
-                    .region(RegionProviderChain::first_try(Region::new(region)))
-                    .load()
-                    .await
+                // This will:
+                // 1. If a region is set, use it
+                // 2. if No region is set, follow the AWS default provider chain
+                //    (https://docs.aws.amazon.com/sdk-for-rust/latest/dg/region.html)
+                // 3. If no region is discovered, hardcode to "us-east-1"
+                //
+                // If there's a region specified we will always prefer that
+                // next, the default provider chain will look at things like AWS_REGION environment
+                // variables, the profile file, etc.
+                //
+                // we provide the hardcoded fallback because a region MUST be set
+                // but, a region being set does _not_ make sense if endpoint_url is set, because
+                // the endpoint URL would include a region.
+                // realistically, endpoint_url is more useful for S3-compatible services
+                // and would not use a separate region in addition to endpoint_url.
+                config_provider = config_provider.region(
+                    RegionProviderChain::first_try(region.map(Region::new))
+                        .or_default_provider()
+                        .or_else(Region::new("us-east-1")),
+                );
+                if let Some(url) = endpoint_url {
+                    config_provider = config_provider.endpoint_url(url)
+                };
+                config_provider.load().await
             });
 
-        let client = s3::client::Client::new(&config);
+        let s3_config = aws_sdk_s3::config::Builder::from(&config)
+            .force_path_style(force_path_style)
+            .build();
+        let client = aws_sdk_s3::Client::from_conf(s3_config);
         Ok(Self { client, rt, bucket })
     }
 
@@ -379,6 +404,14 @@ mod tests {
     ///  * `AWS_TEST_ACCESS_KEY_ID` / `AWS_TEST_SECRET_ACCESS_KEY` - credentials for access to the
     ///    bucket.
     ///
+    /// Additionally, the following environment variables are optional and control
+    /// the created S3 client
+    /// * `AWS_TEST_ENDPOINT_URL` - endpoint URL to use, potentially for an S3-compatible API,
+    ///    or potentially just to ensure this feature works with AWS (e.g. s3.us-east-1.amazonaws.com)
+    /// * `AWS_TEST_FORCE_PATH_STYLE` - if set to "1" or "true", uses "path-style" S3
+    ///   urls ($AWS_TEST_ENDPOINT_URL/$AWS_TEST_BUCKET vs $AWS_TEST_BUCKET.$AWS_TEST_ENDPOINT_URL)
+    ///
+    ///
     /// Set up the bucket with a lifecyle policy to delete objects with age > 1 day. While passing
     /// tests should correctly clean up after themselves, failing tests may leave objects in the
     /// bucket.
@@ -403,14 +436,22 @@ mod tests {
             return None;
         };
 
+        let endpoint_url = std::env::var("AWS_TEST_ENDPOINT_URL").ok();
+
+        let force_path_style = std::env::var("AWS_TEST_FORCE_PATH_STYLE")
+            .map(|f| f == "1" || f == "true")
+            .unwrap_or(false);
+
         Some(
             AwsService::new(
-                region,
+                Some(region),
                 bucket,
                 AwsCredentials::AccessKey {
                     access_key_id,
                     secret_access_key,
                 },
+                endpoint_url,
+                force_path_style,
             )
             .unwrap(),
         )
--- a/src/server/config.rs
+++ b/src/server/config.rs
@@ -79,7 +79,7 @@ pub enum ServerConfig {
     #[cfg(feature = "server-aws")]
     Aws {
         /// Region in which the bucket is located.
-        region: String,
+        region: Option<String>,
         /// Bucket in which to store the task data.
         ///
         /// This bucket must not be used for any other purpose. No special bucket configuration is
@@ -90,6 +90,12 @@ pub enum ServerConfig {
         /// Private encryption secret used to encrypt all data sent to the server.  This can
         /// be any suitably un-guessable string of bytes.
         encryption_secret: Vec<u8>,
+        // endpoint_url is an optional URL to specify the hostname of an s3-compatible service.
+        // When endpoint_url is used, region is ignored by the underlying S3 client.
+        endpoint_url: Option<String>,
+        // force_path_style is used to force the S3 client to use path-style URLs instead of
+        // virtual-hosted-style (subdomain) URLs for the bucket.
+        force_path_style: bool,
     },
 }
 
@@ -120,8 +126,10 @@ impl ServerConfig {
                 bucket,
                 credentials,
                 encryption_secret,
+                endpoint_url,
+                force_path_style,
             } => Box::new(CloudServer::new(
-                AwsService::new(region, bucket, credentials)?,
+                AwsService::new(region, bucket, credentials, endpoint_url, force_path_style)?,
                 encryption_secret,
             )?),
         })
--- a/src/server/config.rs
+++ b/src/server/config.rs
@@ -79,6 +79,17 @@ pub enum ServerConfig {
     #[cfg(feature = "server-aws")]
     Aws {
         /// Region in which the bucket is located.
+        /// If `None`, the default region is used.
+        /// The `default` region is based on the AWS SDK
+        /// - <https://docs.aws.amazon.com/sdk-for-rust/latest/dg/region.html>
+        ///
+        /// following, in order:
+        /// 1. `AWS_REGION` environment variable,
+        /// 2. `AWS_CONFIG_FILE` environment variable and the `region` in that file
+        /// 3. `AWS_PROFILE` variable and the region for that file in the config file
+        /// 4. The instance profile if running in an AWS compute environment
+        ///
+        /// Failing all of those, we will default to `us-east-1`.
         region: Option<String>,
         /// Bucket in which to store the task data.
         ///
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2664,7 +2664,7 @@ dependencies = [
 
 [[package]]
 name = "taskchampion"
-version = "2.0.4-pre"
+version = "3.0.4-pre"
 dependencies = [
  "anyhow",
  "aws-config",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "taskchampion"
-version = "2.0.4-pre"
+version = "3.0.4-pre"
 authors = ["Dustin J. Mitchell <dustin@mozilla.com>"]
 description = "Personal task-tracking"
 homepage = "https://gothenburgbitfactory.github.io/taskchampion/"


src/server/config.rs
```diff
@@ -79,7 +79,7 @@ pub enum ServerConfig {
     #[cfg(feature = "server-aws")]
     Aws {
         /// Region in which the bucket is located.
```
djmitche, line 82: Let's add some docs here as to when this is and is not required.

src/server/cloud/aws.rs
```diff
@@ -92,13 +94,39 @@ impl AwsService {
                         // Just use the default.
                     }
                 }
+                // This will:
+                // 1. If a region is set, use it
+                // 2. if No region is set, follow the AWS deafult provider chain
```
djmitche:
```suggestion
                // 2. if No region is set, follow the AWS default provider chain
```

src/server/config.rs
travisby: How does the current (code) documentation feel here?

travisby:
I'm slowly getting my local rust development environment up to snuff, running docs and tests locally :D.

---

I fixed the _existing_ tests just with skipping the new optional fields.

What would good tests here look like?  Do we want a series of tests that require a real, running S3-compatible storage the way we have the AWS tests?  Is there a nice rust mechanism ("features" maybe?) for running the same tests on different backends?

... or we could keep using the existing AWS implementation, and just pass in the endpoint_url as s3.us-east-1.amazonaws.com :thinking: (although that wouldn't be a _great_ test, since if endpoint_url wasn't piped through correctly, it would _still_ work).

---

For this PR, should I be doing anything for the failed semver check?  I'm pretty sure we expected that in #546, mentioning that in #599 we were already planning a breaking change.

src/server/cloud/aws.rs
```diff
@@ -92,13 +94,36 @@ impl AwsService {
...
-                    .load()
-                    .await
+                // This will:
+                // 1. If a region is set, use it
+                // 2. if No region is set, follow the AWS default provider chain
+                //    (https://docs.aws.amazon.com/sdk-for-rust/latest/dg/region.html)
+                // 3. If no region is discovered, hardcode to "us-east-1"
+                //
+                // If there's a region specified we will always prefer that
+                // next, the default provider chain will look at things like AWS_REGION environment
```
djmitche:
```suggestion
                // If there's a region specified we will always prefer that.
                // Next, the default provider chain will look at things like AWS_REGION environment
```

djmitche (commented):
> I'm slowly getting my local rust development environment up to snuff, running docs and tests locally :D.
> 
> I fixed the _existing_ tests just with skipping the new optional fields.

Nice, and that also verifies that existing users have a simple migration path.

> What would good tests here look like? Do we want a series of tests that require a real, running S3-compatible storage the way we have the AWS tests? Is there a nice rust mechanism ("features" maybe?) for running the same tests on different backends?

For the moment, I think we can document your manual testing against a specific backend, with an issue filed to automate that testing. I think we will want to document which backends we have tested with, and where possible automate that testing -- but some of those backends are probably too complicated to set up.

> ... or we could keep using the existing AWS implementation, and just pass in the endpoint_url as s3.us-east-1.amazonaws.com 🤔 (although that wouldn't be a _great_ test, since if endpoint_url wasn't piped through correctly, it would _still_ work).

This is a good idea just to test the codepaths.

> For this PR, should I be doing anything for the failed semver check? I'm pretty sure we expected that in #546, mentioning that in #599 we were already planning a breaking change.

Yes, you can bump the version in `Cargo.toml`:
```
version = "4.0.0-pre"
```

travisby:
> For the moment, I think we can document your manual testing against a specific backend

I figured out how to use the `cargo test` with all of this!  See results/analysis below!

Step up from "manual test" of `task sync` but still below the full automated run on CI :) 

> This is a good idea just to test the codepaths.

Done below!

> Yes, you can bump the version in Cargo.toml:

Was 4.x intentional?  It looks like 3.x was the next bump!  I went for that, but can happily change up to 4.


---

Alright, good and bad!

I've actually updated those manual tests to allow specifying our two new parameters - endpoint url and force path style

Here's what works:

* `AWS_TEST_REGION=us-east-2 AWS_TEST_BUCKET=test-travis-a96d AWS_TEST_ACCESS_KEY_ID=travis AWS_TEST_SECRET_ACCESS_KEY=hunter2 AWS_TEST_FORCE_PATH_STYLE=true AWS_TEST_ENDPOINT_URL=https://s3.us-east-2.amazonaws.com cargo test`
* `AWS_TEST_REGION=us-east-2 AWS_TEST_BUCKET=test-travis-a96d AWS_TEST_ACCESS_KEY_ID=travis AWS_TEST_SECRET_ACCESS_KEY=hunter2 AWS_TEST_FORCE_PATH_STYLE=true cargo test`
* `AWS_TEST_REGION=us-east-2 AWS_TEST_BUCKET=test-travis-a96d AWS_TEST_ACCESS_KEY_ID=travis AWS_TEST_SECRET_ACCESS_KEY=hunter2 cargo test`

---

If we modify the code to make AWS_TEST_REGION optional, we would still get an error if we did not specify it:
```rust
Caused by:
    Error { code: "PermanentRedirect", message: "The bucket you are attempting to access must be addressed using the specified endpoint. Please send all future requests to this endpoint.", s3_extended_request_id: "sIBjQDiUbiaghYjxvN1sY4FHbz6kb3Z071s6u0+8lrTX29KLNOkyL1lx+otmOBw5ECKgUBo6ynQ=", aws_request_id: "GB3PYDCYZGCRFZ47" })
```

.. if my AWS_PROFILE is setup to use that region (or just the aws-sdk supported env var `AWS_REGION`) then it does work.

I _think_ that's OK

---

Now, what's worse is that despite manual testing being successful (`task sync` working) when using my ceph gateway, _that_ fails on one test:

```sh
AWS_TEST_REGION=fake AWS_TEST_BUCKET=task-test-be13a225-d157-4488-8c56-65cbb9a7a522 AWS_TEST_ACCESS_KEY_ID=travis AWS_TEST_SECRET_ACCESS_KEY=hunter2 AWS_TEST_ENDPOINT_URL=https://mydomain AWS_TEST_FORCE_PATH_STYLE=true cargo test
....
failures:

---- server::cloud::aws::tests::compare_and_swap_matches stdout ----

thread 'server::cloud::aws::tests::compare_and_swap_matches' (252165) panicked at src/server/cloud/test.rs:189:5:
assertion failed: svc.compare_and_swap(&pfx("testy"), Some(b"foo2".to_vec()), b"bar".to_vec())?
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace


failures:
    server::cloud::aws::tests::compare_and_swap_matches
```

(the reason my manual testing worked.. I only added new tasks, I don't think I modified an existing one when I attempted to re-sync?)

----

After a lot of debugging this morning I've gotten it down to:
* PreconditionFailed -- this means the `if_match(etag)` failed.

And, here is what's interesting. When I printed out the etag with some printf-style debugging, I noticed it was printing `The etag is "myetaghere"`

 I setup two aliases, `aaws` that uses AWS s3, and `caws` that uses my ceph object store:

```sh
aaws s3api put-object --bucket test-travis-a96d --key foo --body file2 | jq -r '.ETag'
"b1946ac92492d2347c6235b4d2611184"
caws s3api put-object --bucket task-test-be13a225-d157-4488-8c56-65cbb9a7a522 --key foo --body file2 | jq -r '.ETag'
"b1946ac92492d2347c6235b4d2611184"
```

Note that the ETag value has quotes in it!

OK, let's test that out:

```bash
$ aaws s3api put-object --bucket test-travis-a96d --key foo --body file --if-match '"b1946ac92492d2347c6235b4d2611184"'
{
    "ETag": "\"a10edbbb8f28f8e98ee6b649ea2556f4\"",
    "ChecksumCRC64NVME": "NiebZSsdOpc=",
    "ChecksumType": "FULL_OBJECT",
    "ServerSideEncryption": "AES256"
}
$ caws s3api put-object --bucket task-test-be13a225-d157-4488-8c56-65cbb9a7a522 --key foo --body file --if-match '"b1946ac92492d2347c6235b4d2611184"'

argument of type 'NoneType' is not iterable
```

And let's try if we strip the quotes:

```bash
$ aaws s3api put-object --bucket test-travis-a96d --key foo --body file2 | jq -r '.ETag'
"b1946ac92492d2347c6235b4d2611184"
$ caws s3api put-object --bucket task-test-be13a225-d157-4488-8c56-65cbb9a7a522 --key foo --body file2 | jq -r '.ETag'
"b1946ac92492d2347c6235b4d2611184"
$ aaws s3api put-object --bucket test-travis-a96d --key foo --body file --if-match 'b1946ac92492d2347c6235b4d2611184'
{
    "ETag": "\"a10edbbb8f28f8e98ee6b649ea2556f4\"",
    "ChecksumCRC64NVME": "NiebZSsdOpc=",
    "ChecksumType": "FULL_OBJECT",
    "ServerSideEncryption": "AES256"
}
$ caws s3api put-object --bucket task-test-be13a225-d157-4488-8c56-65cbb9a7a522 --key foo --body file --if-match 'b1946ac92492d2347c6235b4d2611184'
{
    "ETag": "\"a10edbbb8f28f8e98ee6b649ea2556f4\""
}
```

---

So, the AWS API supports using the quoted value from the ETag response directly (which includes quotes intentionally.  There's some previous aws sdk issues on GH about it I can add later), but at least my Ceph object gateway does not!

----

I'm not sure if the right answer is for us to strip quotes here, because it _appears_ to work with AWS S3 without them (do they document anywhere that it should work?), add _more_ configuration options to support this for s3-"compatible" apis, or submit an upstream ticket to ceph (I probably should do that either way).  We also can generate the etag from the file contents instead of relying on an extra get_object call + comparing content.  The ETag that s3 returns is just an md5sum of the file contents:

```sh
md5sum < file2
b1946ac92492d2347c6235b4d2611184  -
```

I think I want to check if one more s3-compatible API has this same problem

[edit] https://tracker.ceph.com/issues/64439 rhymes with our current problem, but I can't quite say for sure if it's the same

travisby:
Testing minio fails, but on a different test :roll_eyes: 

```sh
failures:

---- server::cloud::aws::tests::compare_and_swap_disappears stdout ----
deleting object fcacc68a90e3426b81a400a551299155-racing-delete

thread 'server::cloud::aws::tests::compare_and_swap_disappears' (254663) panicked at src/server/cloud/test.rs:250:5:
assertion failed: !svc.compare_and_swap(&pfx("racing-delete"), Some(b"foo1".to_vec()),
            b"bar".to_vec())?
```

(it succeeded `server::cloud::aws::tests::compare_and_swap_matches` though)

travisby:
> We also can generate the etag from the file contents instead of relying on an extra get_object call + comparing content. The ETag that s3 returns is just an md5sum of the file contents:

This patch switches to calculating the md5sum, and avoiding the extra `get_object` call.  It works on both s3 and ceph (because we're not adding in the extra quotes) (still fails on minio with the same error as before.. have not dug in yet)

```diff
diff --git a/Cargo.lock b/Cargo.lock
index 26cf0ae9d..b966e1678 100644
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1639,6 +1639,12 @@ dependencies = [
  "digest",
 ]
 
+[[package]]
+name = "md5"
+version = "0.8.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "ae960838283323069879657ca3de837e9f7bbb4c7bf6ea7f1b290d5e9476d2e0"
+
 [[package]]
 name = "memchr"
 version = "2.7.4"
@@ -2675,6 +2681,7 @@ dependencies = [
  "flate2",
  "google-cloud-storage",
  "log",
+ "md5",
  "pretty_assertions",
  "proptest",
  "ring",
diff --git a/Cargo.toml b/Cargo.toml
index 6faf0908b..4b8dd0726 100644
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -67,6 +67,7 @@ thiserror = "2.0"
 ureq = { version = "^2.12.1", features = ["tls"], optional = true }
 uuid = { version = "^1.16.0", features = ["serde", "v4"] }
 url = { version = "2", optional = true }
+md5 = "0.8.0"
 
 [dev-dependencies]
 proptest = "^1.7.0"
diff --git a/src/server/cloud/aws.rs b/src/server/cloud/aws.rs
index 7f74c80f3..6ede4888b 100644
--- a/src/server/cloud/aws.rs
+++ b/src/server/cloud/aws.rs
@@ -226,35 +226,8 @@ impl Service for AwsService {
     ) -> Result<bool> {
         self.block_on(async {
             validate_object_name(name);
-            let get_res = if_key_exists(
-                self.client
-                    .get_object()
-                    .bucket(self.bucket.clone())
-                    .key(name)
-                    .send()
-                    .await
-                    .map_err(aws_err),
-            )?;
 
-            // Check the expectation and gather the e_tag for the existing value.
-            let e_tag;
-            if let Some(get_res) = get_res {
-                // If a value was not expected but one exists, that expectation has not been met.
-                let Some(existing_value) = existing_value else {
-                    return Ok(false);
-                };
-                e_tag = get_res.e_tag.clone();
-                let body = get_body(get_res).await?;
-                if body != existing_value {
-                    return Ok(false);
-                }
-            } else {
-                // If a value was expected but none exists, that expectation has not been met.
-                if existing_value.is_some() {
-                    return Ok(false);
-                }
-                e_tag = None;
-            };
+            let e_tag = existing_value.map(md5::compute).map(|d| format!("{:x}", d));
 
             // When testing, an object named "$pfx-racing-delete" is deleted between get_object and
             // put_object.
```

(I have no idea if the `format!` is the best way to get an ascii representation of a `Digest` but it's the best I could find online/figure out)

---

Thoughts?  I'm not sure how I feel about a change like this to get around another service's bug :|, but on the bright side we reduce an extra API call?

travisby:
The reason minio does not pass the racing delete test.. 

```sh
$ maws s3api head-object --bucket test --key foo

An error occurred (404) when calling the HeadObject operation: Not Found
$ maws s3api put-object --bucket test --key foobar --body file2 --if-match b1946ac92492d2347c6235b4d2611184
{
    "ETag": "\"b1946ac92492d2347c6235b4d2611184\"",
    "ChecksumCRC64NVME": "akP7S61aVgc=",
    "ChecksumType": "FULL_OBJECT"
}
```

minio _ignores_ the `if-match` when the object does not yet exist :dizzy_face:

travisby:
We can't/shouldn't assume the ETag is an md5sum: https://docs.aws.amazon.com/AmazonS3/latest/userguide/checking-object-integrity-upload.html#checking-object-integrity-etag-and-md5

(If you use a bucket w/ KMS encryption, it won't be).

djmitche:
It seems like Minio doesn't provide a way to reliably compare-and-swap a value for the absence of a value (that is, to prevent two racing processes from succeeding at putting objects when none exists). So that's not actually enough to support Taskchampion.

I think the Ceph issue can be worked around by just stripping the quotes: both services include quotes in responses, and both services will accept an unquoted string in a request. Or have I missed a further issue there?

travisby:
Exciting news!  minio was fixed in https://github.com/minio/minio/issues/21526#issuecomment-3263703407

And as of their version: `RELEASE.2025-09-07T16-13-09Z` all taskchampion tests pass!

`AWS_TEST_REGION=fake AWS_TEST_BUCKET=<directory in my minio path> AWS_TEST_ACCESS_KEY_ID=minioadmin AWS_TEST_SECRET_ACCESS_KEY=minioadmin AWS_TEST_ENDPOINT_URL=http://localhost:9000 AWS_TEST_FORCE_PATH_STYLE=true cargo test`

----

On the ceph front, :+1: stripping quotes seems to work and we _could_ get tests to pass with: `e_tag = e_tag.replace('"', "");` (this works with s3, minio, _and_ ceph)


That being said, 
* https://datatracker.ietf.org/doc/html/rfc7232#section-3.2:~:text=An%20entity%2Dtag%20consists%20of%20an%20opaque%20quoted%20string
* https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/If-None-Match#:~:text=If%2DNone%2DMatch%3A%20%22%3Cetag_value%3E%22%2C%20%22%3Cetag_value%3E%22%2C%20%E2%80%A6

It looks like stripping quotes would technically be non-conformant with the spec, as they consider the quotes _part_ of the etag.  

We can't say for sure if this could break other implementations (there's others like [Heztner's](https://www.hetzner.com/storage/object-storage/) that I won't be able to test).

---

Do we want to intentionally release something non-conformant, with it in mind that it works in practice for the implementations we see?  And if a new implementation comes along and we have to choose between breaking ceph that is non-compliant and supporting $newthing that is compliant.. make the choice then? (Or flagg-ify it?)

djmitche:
https://tracker.ceph.com/issues/64439 does seem like the same issue, and I think that's a bug in Ceph. We should probably try harder to get that fixed and only build in a workaround if both (a) someone really wants Ceph support and (b) Ceph provides a reasonable justification for not fixing the bug.  I realize I'm contradicting an earlier comment of mine, sorry :)

But for the moment it looks like we could advertise support for minio, with the minimum version you specified -- that seems like a great thing to get landed now. We can include a note in the docs that Ceph support is waiting on a fix for the linked bug.

djmitche:
We are getting closer to a 3.0 release (a week or two, maybe?). Do you think we can get this included?

djmitche:
I merged this with main. One issue I found was that the default region search included consulting IMDS, which requires an HTTP client and did not use the configured (reqwest-based) HTTP client. I don't think using IMDS is that critical for our use-case, so I just omitted that. WDYT?

src/server/cloud/aws.rs
```diff
@@ -102,14 +106,39 @@ impl AwsService {
             }
         }
 
+        eprintln!("uh, here");
```
ryneeverett: :wave:

ryneeverett (approved):

src/server/cloud/aws.rs
djmitche: The printf-debugging of WASM tests is rough, sorry!

PR #602, opened 2025-08-10
Fix rustdoc not-a-link

travisby:
In #601 we noticed a new failure on rustdoc.

It wasn't happening locally for me, so I went and checked with nightly (it looks like our `rustdoc` action uses nightly) + the recent changelog for rust.  This landed just a few days ago: rust-lang/rust/pull/132748 and seems very suspiciously the culprit.

I don't fully understand what our intention was here, but I swapped to _no_ backticks based on this comment:

```rust
// If there's no backticks, be lenient and revert to the old behavior.
```

I have no idea if this was the right solution, besides the fact CI passes :rofl: .

### DIFF
--- a/src/replica.rs
+++ b/src/replica.rs
@@ -306,7 +306,7 @@ impl Replica {
 
     /// Create a new task.
     ///
-    /// Use ['Uuid::new_v4`] to invent a new task ID, if necessary. If the task already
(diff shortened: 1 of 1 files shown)


djmitche (approved):
I don't totally understand either, but if the robots are OK with it then I am too.

PR #604, opened 2025-08-18
Refactor to closure-based tx signature

geofflittle:
This pull request refactors the storage layer to use a new, closure-based transaction signature. The is to ultimately align our transaction handling with the what's required by the async database lib `tokio_rusqlite`, which is a step toward making this library async.

This PR depends on #600.

### DIFF
--- a/src/replica.rs
+++ b/src/replica.rs
@@ -2,6 +2,8 @@ use crate::depmap::DependencyMap;
 use crate::errors::Result;
 use crate::operation::{Operation, Operations};
 use crate::server::Server;
+#[cfg(test)]
+use crate::storage::InMemoryStorage;
 use crate::storage::{Storage, StorageTxn, TaskMap};
 use crate::task::{Status, Task};
 use crate::taskdb::TaskDb;
@@ -35,15 +37,15 @@ use uuid::Uuid;
 # #[cfg(feature = "storage-sqlite")]
 # {
 # use taskchampion::chrono::{TimeZone, Utc};
-# use taskchampion::{storage::AccessMode, Operations, Replica, Status, StorageConfig, Uuid};
+# use taskchampion::{storage::AccessMode, Operations, Replica, Status, Uuid, storage::SqliteStorage};
 # use tempfile::TempDir;
 # fn main() -> anyhow::Result<()> {
 # let tmp_dir = TempDir::new()?;
-# let mut replica = Replica::new(StorageConfig::OnDisk {
-#   taskdb_dir: tmp_dir.path().to_path_buf(),
-#   create_if_missing: true,
-#   access_mode: AccessMode::ReadWrite,
-# }.into_storage()?);
+# let mut replica: Replica<SqliteStorage> = Replica::new(SqliteStorage::new(
+#   tmp_dir.path().to_path_buf(),
+#   AccessMode::ReadWrite,
+#   true,
+# )?);
 // Create a new task, recording the required operations.
 let mut ops = Operations::new();
 let uuid = Uuid::new_v4();
@@ -69,8 +71,8 @@ replica.commit_operations(ops)?;
 /// specifically pending tasks.  These are indexed with small, easy-to-type integers.  Newly
 /// pending tasks are automatically added to the working set, and the working set can be
 /// "renumbered" when necessary.
-pub struct Replica {
-    storage: Box<dyn Storage>,
+pub struct Replica<S: Storage> {
+    storage: S,
 
     taskdb: TaskDb,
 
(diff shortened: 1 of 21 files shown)


src/storage/config.rs
```diff
@@ -10,6 +10,25 @@ pub enum AccessMode {
     ReadWrite,
 }
 
+pub enum StorageEnum {
```
djmitche: IIUC the reason to switch from `dyn Storage` is that with the generic `txn` method `Storage` is no longer dyn-compatible.

The approach of using an enum precludes users of the library from bringing their own storage implementation, though -- and that's something I'd like to keep supporting.

Could we instead make Replica generic over the storage: `struct Replica<S> where S: Storage { .. }`? That would feel more idiomatic.

It might make the FFI a litle more complicated, but we can use this same enum trick at that point, since the FFI only supports a few storage types. In other words the FFI implementation in Taskwarrior will have something like
```rust
enum Replica {
    OnDisk(taskchampion::Replica<SqliteStorage>),
    InMemory(taskchampion::Replica<InMemoryStorage>)
}
```
and wrap each method call in a `match` just like you've done here.

src/replica.rs
```diff
@@ -439,10 +448,10 @@ impl Replica {
     /// This method only supports reversing operations if they precisely match local operations
     /// that have not yet been synchronized, and will return `false` if this is not the case.
     pub fn commit_reversed_operations(&mut self, operations: Operations) -> Result<bool> {
-        if !self
-            .taskdb
-            .commit_reversed_operations(self.storage.txn()?.as_mut(), operations)?
-        {
+        let is_rev = self
```
djmitche: Perhaps `success` or `succeeded` would be better variable names

src/storage/config.rs
geofflittle: Got it!

I made this change and added some TryFrom helpers in StorageConfig, lmk if you don't prefer those.

src/storage/mod.rs
```diff
@@ -22,10 +22,15 @@ mod config;
 #[cfg(feature = "storage-sqlite")]
 pub(crate) mod sqlite;
 
+#[cfg(feature = "storage-sqlite")]
+pub use sqlite::SqliteStorage;
```
djmitche, line 25: I was going to object to making these types public, as they have previously been accessed via `StorageConfig::into_storage`. However, that function returns `Box<dyn Storage>` and a boxed storage is no longer useful -- so I think `StorageConfig` itself is no longer useful.

All of which to say, I think this is fine, but let's also remove the StorageConfig type at the same time.

src/storage/config.rs
```diff
@@ -30,20 +33,48 @@ pub enum StorageConfig {
     InMemory,
 }
 
-impl StorageConfig {
-    pub fn into_storage(self) -> Result<Box<dyn Storage>> {
-        Ok(match self {
+#[cfg(feature = "storage-sqlite")]
+impl TryFrom<StorageConfig> for SqliteStorage {
```
djmitche: Relating to my comment in `src/storage/mod.rs`, this is pretty awkward: this is taking an enum type and returning a concrete type _if_ the correct enum variant is present, otherwise producing a runtime error. This should be a build-time error, and I think just instantiating the `{Sqlite,InMemory}Storage` types directly as arguments to `Replica::new` is the simplest and best way to accomplish that.

It means the `StorageConfig` type and this module can go away entirely!

djmitche (commented):
Looks good aside from the suggestions below. I realize this suggestion will end up rewriting just about every test!

So, before doing that, it might make sense to verify that this callback-style approach still works in an async context. You might just temporarily update the "real" code and not the tests and see if `cargo build` passes. I'm worried that things like `Send` or `'static` will make this really challenging.

geofflittle:
> it might make sense to verify that this callback-style approach still works in an async context

I think what you're looking for is present in #607!

djmitche:
Oh, yes -- carry on, then! I just wanted to avoid a lot of work leading to a dead end :)

geofflittle:
I'm with ya, thank you for checking!

I'll make those StorageConfig changes shortly!

src/storage/config.rs
geofflittle: Removed!

src/replica.rs
```diff
@@ -92,10 +94,8 @@ impl Replica {
     }
 
     #[cfg(test)]
-    pub fn new_inmemory() -> Replica {
-        use crate::StorageConfig;
-
-        Replica::new(StorageConfig::InMemory.into_storage().unwrap())
+    pub(crate) fn new_inmemory() -> Replica<InMemoryStorage> {
```
djmitche, line 97: Let's make this an easy followup PR: this function isn't necessary and is only used in tests anyway, so its uses can be replaced by its body.

djmitche (approved):
Nice work!!

PR #607, opened 2025-08-20
Refactor: Convert Replica and Storage layers to async

geofflittle:
This change depends on #604.

The `Replica` and `Storage` APIs have been converted to be async. This makes the library non-blocking and suitable for use in WASM and other environments.

Key changes include:
* Most methods on `Replica` are now `async` and must be awaited.
* The `Storage` trait is now an `async_trait`, with its `txn` method becoming async.
* `SqliteStorage` is reimplemented using `tokio-rusqlite` to provide an async interface to the database. This required downgrading the `rusqlite` dependency to a compatible version.
* `InMemoryStorage` now uses a `Mutex` to ensure thread-safe access in async contexts.
* The `TaskDb` struct has been refactored into a stateless "namespace", removing it as a field from `Replica` and simplifying state management.
* `ServerConfig::into_server` now returns an `Arc<Mutex<dyn Server + Send>>` to allow servers to be shared safely across async tasks.
* All tests, examples, and internal call sites have been updated for the new async API, using `#[tokio::test]` and `await`.

### DIFF
--- a/src/storage/mod.rs
+++ b/src/storage/mod.rs
@@ -10,6 +10,7 @@ traits defined here and pass the result to [`Replica`](crate::Replica).
 
 use crate::errors::Result;
 use crate::operation::Operation;
+use async_trait::async_trait;
 use std::collections::HashMap;
 use uuid::Uuid;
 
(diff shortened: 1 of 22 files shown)


src/storage/mod.rs
```diff
@@ -153,9 +154,11 @@ pub trait StorageTxn {
...
-pub trait Storage {
-    /// Begin a transaction
-    fn txn<F, R>(&mut self, f: F) -> Result<R>
+#[async_trait]
+pub trait Storage: Send + Sync {
+    /// Begin an async transaction
+    async fn txn<F, R>(&self, f: F) -> Result<R>
     where
-        F: for<'a> FnOnce(&'a mut (dyn StorageTxn + 'a)) -> Result<R>;
+        F: for<'a> FnOnce(&'a mut (dyn StorageTxn + 'a)) -> Result<R> + Send + 'static,
```
djmitche, line 174: Woah, I didn't realize tokio_rusqlite::Connection::calll took a _sync_ callback to invoke in a background thread. That's pretty wild! It's worth dwelling on this in the doc comment for the trait, since authors of future implementations of this trait (e.g., postgres) may need to emulate this.

djmitche (commented):
I apologize this took so long to review, but I'm quite happy with the result!

src/storage/mod.rs
geofflittle: Done!

djmitche (approved):
Awesome, thank you!!

Issue #613, opened 2025-08-28
Make simple property getters/setters available in TaskData


martiege:
Hello! 

I'm continuing to work on my small app, and as I've always found the "urgency" number pretty useful, I figured I'd also show this off in the app, so I've started to rewrite that part of the code.

The issue I came across, is that when I want to implement the urgency function for the scheduled property, I have to parse this value myself, instead of e.g. using the `get_due` method. 

For now, I've essentially recreated the `get_timestamp` function from task.rs, which doesn't feel very satisfying. 

I'm guessing that it isn't logical to have methods for all possible properties, but perhaps it could make sense to put some of the utility functions either in the TaskData structure, or in some separate utility module, so that the timestamp-functionality can be reused for the other relevant properties when necessary?

djmitche:
That makes a lot of sense. Some of them could probably just be made `pub`? I'd be happy to look at a PR for this.

djmitche:
I've updated the issue subject. As I understand it, the idea here is to make the "simple" getters and setters like `get_status` or `set_wait` available on TaskData.

I suspect we could avoid some duplication by `impl Deref<TaskData> for Task`, so that TaskData methods can be called on a Task. Then, we can simply move those method implementations to the TaskData type.

PR #614, opened 2025-08-29
changing timestamp-utility functions to pub

martiege:
As suggested in issue #613, I've moved the timestamp utilities public. Also physically moved the functions outside the comment for utility functions for clarity. Will allow the same functionality to be used for other timestamp data in the task, such as `scheduled`.

### DIFF
--- a/src/task/task.rs
+++ b/src/task/task.rs
@@ -546,16 +546,7 @@ impl Task {
         self.set_value(key, None, ops)
     }
 
-    // -- utility functions
-
-    fn is_known_key(key: &str) -> bool {
-        Prop::from_str(key).is_ok()
-            || key.starts_with("tag_")
-            || key.starts_with("annotation_")
-            || key.starts_with("dep_")
-    }
-
-    fn get_timestamp(&self, property: &str) -> Option<Timestamp> {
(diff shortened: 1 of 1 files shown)


src/task/task.rs
```diff
@@ -546,16 +546,7 @@ impl Task {
...
-
-    fn is_known_key(key: &str) -> bool {
-        Prop::from_str(key).is_ok()
-            || key.starts_with("tag_")
-            || key.starts_with("annotation_")
-            || key.starts_with("dep_")
-    }
-
-    fn get_timestamp(&self, property: &str) -> Option<Timestamp> {
+    pub fn get_timestamp(&self, property: &str) -> Option<Timestamp> {
```
djmitche, line 553: Now that it's pub, this should get a documentation comment (and `put_timestamp` too).

djmitche (commented):
Apologies for the delay here!

src/task/task.rs
martiege: Added some now, can you check that they seem reasonable? I've tried to follow the same style as the other documentation comments.

djmitche (approved):
Thank you!

PR #617, opened 2025-09-09
feat: Add WASM Support

geofflittle:
This PR introduces support for compiling to the `wasm32-unknown-unknown` target, enabling its use in WASM environments.

To achieve this, several changes were made:
* **WASM-Specific Dependencies:** `Cargo.toml` now uses target-specific dependencies. For `wasm32`, `tokio` is configured with a single-threaded runtime and `uuid` to use JS-native random number generation. Non-WASM targets retain the multi-threaded `tokio` runtime.
* **Feature Flag Refactoring:** A new `sync-api` private feature was introduced to gate synchronization logic that is not WASM-compatible. Existing sync features now depend on this new feature flag. This allows building the core library without the networking and filesystem dependencies required for server sync.
* **In-Memory Storage:** A new `storage-inmemory` feature has been added. A future PR will add this functionality to provide a storage backend that is compatible with WASM environments.
* **CI Workflow:** A new GitHub Action, `build-wasm`, has been added to continuously verify that the crate builds successfully for the `wasm32` target.
* **Documentation:** The `README.md` has been updated with a new section explaining how to build the crate for WASM, highlighting the need to disable default features and select compatible ones.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,20 +20,24 @@ resolver = "2"
 [features]
 default = ["sync", "bundled", "storage"]
 
+# (private) Support for the sync API
+sync-api = []
 # Support for all sync solutions
-sync = ["server-sync", "server-gcp", "server-aws", "server-local"]
+sync = ["sync-api", "server-sync", "server-gcp", "server-aws", "server-local"]
 # Support for sync to a server
-server-sync = ["encryption", "dep:ureq", "dep:url"]
+server-sync = ["sync-api", "encryption", "dep:ureq", "dep:url"]
 # Support for sync to GCP
-server-gcp = ["cloud", "encryption", "dep:google-cloud-storage"]
+server-gcp = ["sync-api", "cloud", "encryption", "dep:google-cloud-storage"]
 # Support for sync to AWS
-server-aws = ["cloud", "encryption", "dep:aws-sdk-s3", "dep:aws-config", "dep:aws-credential-types"]
+server-aws = ["sync-api", "cloud", "encryption", "dep:aws-sdk-s3", "dep:aws-config", "dep:aws-credential-types"]
 # Suppport for sync to another SQLite database on the same machine
-server-local = ["storage-sqlite"]
+server-local = ["sync-api", "storage-sqlite"]
 # Support for all task storage backends
 storage = ["storage-sqlite"]
 # Support for SQLite task storage
 storage-sqlite = ["dep:rusqlite", "dep:tokio-rusqlite"]
+# Support for the in-memory storage backend (WASM-compatible)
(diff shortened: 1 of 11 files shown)


tests/update-and-delete-sync.rs
```diff
@@ -1,23 +1,22 @@
+#![cfg(feature = "server-local")]
```
djmitche, line 1: Oh, that's cool!

Cargo.toml
```diff
@@ -42,6 +44,8 @@ cloud = []
 bundled = ["rusqlite/bundled", "tokio-rusqlite/bundled"]
 # use native CA roots, instead of bundled
 tls-native-roots = ["ureq/native-certs"]
+# Feature for wasm builds. This enables wasm-compatible storage.
+wasm = ["storage-inmemory"]
```
djmitche: I think you'll need `--no-default-features --features wasm` here, but the WASM is implicit from the platform. It might make sense to just document that when building for WASM, only a few features are allowed (among them `storage-inmemory`) and `--no-default-features` will be required. Then users building for WASM can select the features they want from among that set (for example, some might want storage-inmemory, while others might want storage-somebrowserdbthing).

Special bonus points if using other features results in a useful error message!

djmitche (commented):
Looking good!

It will be good to update the GitHub actions to do a WASM build, too, so we don't accidentally regress the various `cfg` conditionals.

Cargo.toml
geofflittle: Got it! I've removed the `wasm` flag and added some treatment in the README.

Re: your suggestion for a better WASM error message, I found that a `build.rs` script can't catch a missing `--no-default-features` flag (i.e., a user building with defaults), because the project's WASM-incompatible dependencies fail to compile *before* `build.rs` runs.

Given this, how do you see our detecting the right situation to display an improved error message?

djmitche: > Given this, how do you see our detecting the right situation to display an improved error message?

Good idea to try to detect in build.rs -- but bummer that it doesn't work. Maybe those special bonus points aren't achievable! No need to spend more time on it.

README.md
```diff
@@ -20,3 +20,12 @@ The Rust API, as defined in [the docs](https://docs.rs/taskchampion/latest/taskc
 
 The Rust API follows semantic versioning.
 As this is still in the `0.x` phase, so breaking changes may occur but will be indicated with a change to the minor version.
+
+## Building for WebAssembly (WASM)
```
djmitche, line 24: Aside from remembering to summarize WASM support in the README when we are done, I think the other "good parts" of this PR have all landed elsewhere, so I'm going to close it up.

PR #619, opened 2025-09-15
Revert async changes

geofflittle:
This PR reverts the recent `async` refactoring.

After further investigation, the initial approach using `tokio_rusqlite` proved to be unworkable (see issue #599). This revert is a clean course correction to clear the way for a more robust solution.

The new implementation, which uses an actor model to wrap the synchronous `SqliteStorage`, is available for review here: ~[link](https://github.com/geofflittle/taskchampion/pull/1)~ #620.


djmitche:
I'm glad to know these will be replaced with something even better :D

PR #620, opened 2025-09-15
refactor: Make storage async

geofflittle:
This PR converts the `Storage` layer and `Replica` API to be fully `async`, making the library non-blocking and compatible with async Rust runtimes.

The synchronous `rusqlite` backend is adapted for the async API by introducing `SqliteStorageActor`. This wrapper uses the actor model, running all database operations on a dedicated background thread and communicating via message passing.

### Breaking API Changes

* **Async Methods:** Most methods on `Replica` and `StorageTxn` are now `async` and must be `.await`ed.
* **`StorageConfig` Removed:** Storage implementations must be constructed directly.
    * **Before:** `Replica::new(StorageConfig::OnDisk { ... }.into_storage()?)`
    * **After:** `Replica::new(SqliteStorageActor::new(...)?)`
* **Generic `Replica`:** The `Replica` struct is now generic over its storage type (`Replica<S: Storage>`) instead of using a trait object.

### DIFF
--- a/src/storage/sqlite/mod.rs
+++ b/src/storage/sqlite/mod.rs
@@ -1,0 +1,428 @@
+use crate::errors::{Error, Result};
+use crate::operation::Operation;
+use crate::storage::config::AccessMode;
+use crate::storage::sqlite::inner::SqliteStorageInner;
+use crate::storage::{Storage, StorageTxn, TaskMap, VersionId};
+use async_trait::async_trait;
+use rusqlite::types::FromSql;
+use rusqlite::ToSql;
+use std::path::Path;
+use std::sync::mpsc;
+use std::thread;
+use tokio::sync::oneshot;
+use uuid::Uuid;
+
+mod inner;
+mod schema;
+
+#[derive(Debug, thiserror::Error, PartialEq, Eq)]
+pub(crate) enum SqliteError {
+    #[error("SQLite transaction already committted")]
+    TransactionAlreadyCommitted,
+    #[error("Task storage was opened in read-only mode")]
+    ReadOnlyStorage,
+}
+
+/// Newtype to allow implementing `FromSql` for foreign `uuid::Uuid`
+pub(crate) struct StoredUuid(pub(crate) Uuid);
+
+/// Conversion from Uuid stored as a string (rusqlite's uuid feature stores as binary blob)
+impl FromSql for StoredUuid {
+    fn column_result(value: rusqlite::types::ValueRef<'_>) -> rusqlite::types::FromSqlResult<Self> {
+        let u = Uuid::parse_str(value.as_str()?)
+            .map_err(|_| rusqlite::types::FromSqlError::InvalidType)?;
+        Ok(StoredUuid(u))
+    }
+}
+
+/// Store Uuid as string in database
+impl ToSql for StoredUuid {
+    fn to_sql(&self) -> rusqlite::Result<rusqlite::types::ToSqlOutput<'_>> {
+        let s = self.0.to_string();
+        Ok(s.into())
+    }
+}
+
+/// An enum for messages sent to the sync thread actor.
+pub(crate) enum ActorMessage {
+    // Transaction control
+    BeginTxn(oneshot::Sender<Result<mpsc::Sender<TxnMessage>>>),
+}
+
+pub(crate) enum TxnMessage {
+    Commit(oneshot::Sender<Result<()>>),
+    Rollback(oneshot::Sender<Result<()>>),
+
+    // Transactional operations
+    GetTask(Uuid, oneshot::Sender<Result<Option<TaskMap>>>),
+    GetPendingTasks(oneshot::Sender<Result<Vec<(Uuid, TaskMap)>>>),
+    CreateTask(Uuid, oneshot::Sender<Result<bool>>),
+    SetTask(Uuid, TaskMap, oneshot::Sender<Result<()>>),
+    DeleteTask(Uuid, oneshot::Sender<Result<bool>>),
+    AllTasks(oneshot::Sender<Result<Vec<(Uuid, TaskMap)>>>),
+    AllTaskUuids(oneshot::Sender<Result<Vec<Uuid>>>),
+    BaseVersion(oneshot::Sender<Result<VersionId>>),
+    SetBaseVersion(VersionId, oneshot::Sender<Result<()>>),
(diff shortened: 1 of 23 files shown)


djmitche (commented):
Just having a quick look through this now, but I like this quite a bit! Putting SQLite in its own thread with an actor pattern is a great idea.

I'm sorry I'm taking so long with reviews. If it's any consolation I'm taking a long time with lots of things these days.

src/storage/sqlite/actor.rs
```diff
@@ -0,0 +1,393 @@
...
+    }
+}
+
+/// SqliteStorageActor is an async actor wrapper for the sync SqliteStorage.
+#[derive(Clone)]
+pub struct SqliteStorageActor {
+    sender: mpsc::Sender<ActorMessage>,
+}
+
+impl SqliteStorageActor {
```
djmitche: Let's shift the naming so that the `pub` name for SQLite storage is just `SqliteStorage` and the fact that there's an actor pattern in use is hidden. It might make a bit more sense to move both types into `src/storage/sqlite/`: right now, it's a bit confusing to see `inmemory.rs` which impl's the StorageTxn type, and `sqlite.rs`, which does not but kind of looks like it does.

src/storage/sqlite/actor.rs
```diff
@@ -0,0 +1,393 @@
...
+                ActorMessage::SetWorkingSetItem(i, u, resp) => {
+                    let _ = resp.send(txn.set_working_set_item(i, u));
+                }
+                ActorMessage::ClearWorkingSet(resp) => {
+                    let _ = resp.send(txn.clear_working_set());
+                }
+                ActorMessage::BeginTxn => {
+                    // This is a logic error. A new transaction was requested
+                    // while one was already in progress.
+                    log::error!("Received BeginTxn while a transaction was already in progress");
```
djmitche: Let's make this a panic -- I think the lifetimes on `storage.txn()` mean that `storage` is borrowed until the resulting transaction is dropped, so this should be impossible.

In fact, as written the channel is used for two different kinds of messages: beginning transactions (BeginTxn) and transaction methods (everything else). We could avoid panics by separating these two kinds into different channels. So the "main" agent channel just accepts `ActorMessage::BeginTxn(oneshot::Sender<Result<mpsc::Sender<TxnMessage>>>)`, then only listens on the corresponding `Receiver<TxnMessage>` until commit or rollback.

Then if, somehow, `storage.txn()` gets called twice concurrently, the second call will send `ActorMessage::BeginTxn` and block waiting for the response, which won't occur untli the first transaction is committed or rolled back -- precisely the sequential semantics the trait requires.

src/storage/sqlite/mod.rs
```diff
@@ -0,0 +1,393 @@
...
+    fn run(&mut self) {
+        // The outer loop waits for a BeginTxn message. If the channel is disconnected,
+        // the thread will exit gracefully.
+        while let Ok(ActorMessage::BeginTxn) = self.receiver.recv() {
+            match self.storage.txn() {
+                Ok(mut txn) => Self::handle_transaction(&self.receiver, &mut txn),
+                Err(e) => {
+                    // If we can't start a transaction, something is wrong.
+                    // Log the error and exit the thread.
+                    log::error!("Could not start SQLite transaction: {e}");
```
djmitche, line 102: This is probably the common place to find an error, e.g., read-only filesystem or corrupt DB. Most uses of this library, such as Taskwarrior, will not typically enable or handle log output, so this error would be invisible to the user.

Could BeginTxn take a oneshot channel to send back an error, too?

src/storage/sqlite/actor.rs
```diff
@@ -0,0 +1,393 @@
...
+                Ok(mut txn) => Self::handle_transaction(&self.receiver, &mut txn),
+                Err(e) => {
+                    // If we can't start a transaction, something is wrong.
+                    // Log the error and exit the thread.
+                    log::error!("Could not start SQLite transaction: {e}");
+                    break;
+                }
+            }
+        }
+        // Exiting the loop means the sender was dropped, so the thread can terminate.
```
djmitche: This isn't quite true -- as written, it could be any of
 - sender was dropped, so `self.receiver.recv` returned an Err
 - `self.receiver.recv` returned Ok but not BeginTxn
 - there was an error starting the transaction

For the first, a clean return makes sense. The second should be a panic, as it should be impossible. And the third needs to end up back with the caller of the `Replica` method as a `taskchampion::Error` as suggested above.

djmitche (commented):
My apologies, again, for taking so long. This is really sophisticated stuff and I definitely needed to have time to concentrate on it!

The comments below are relatively minor -- this looks great, and I'm quite happy with the result!

src/storage/sqlite/actor.rs
geofflittle: Fixed!

src/storage/sqlite/mod.rs
geofflittle: Added the new channel!

src/storage/sqlite/actor.rs
geofflittle: Changed!

src/storage/sqlite/actor.rs
geofflittle: Added the new channel!

src/storage/sqlite/mod.rs
```diff
@@ -0,0 +1,428 @@
...
+    async fn call<R, F>(&self, f: F) -> Result<R>
+    where
+        F: FnOnce(oneshot::Sender<Result<R>>) -> TxnMessage,
+        R: Send + 'static,
+    {
+        let (tx, rx) = oneshot::channel();
+        self.sender
+            .send(f(tx))
+            .map_err(|e| Error::Other(e.into()))?;
+        rx.await?
```
djmitche, line 264: This async call could be cancelled here, and I wanted to think through what would happen in that case.

The `rx` would be dropped meaning an error in the actor thread when the query is complete. But that's ignored with `let _ = ..`. So it goes around the loop again waiting for another message. If the ActorTxn is not dropped, this is fine -- it will either send another query or commit/rollback.

If the ActorTxn is also dropped when this call is cancelled, then its `drop` method will get called, so when the current query completes, the actor will see and ignore an error, then go around the loop again and receive the Rollback message.

All of that is well and good, but I figured it'd be worth recording the thought process!

djmitche (approved):
Awesome stuff!

PR #625, opened 2025-10-04
Update README.md to reflect non 0.x status

miguelpduarte:
Taskchampion has since left `0.x` versions, so the README can be updated to reflect that 🙂


djmitche (approved):
Good spot, thank you!

PR #626, opened 2025-10-04
Add serde Serialize and Deserialize to Task and TaskData

### LINKED ISSUE #624: Expose a serde-Serialize/Deserialize-able Task representation
I'm building a project with `taskchampion` and just absentmindedly tried to pass around a `Task` over a serialization boundary (in this case, JSON), when I noticed that it does not implement either of the typical `serde` traits (`Serialize` and `Deserialize`).

Would it make sense to either provide a "view" over a `Task` (e.g. `ReadOnlyTask` or similar) that would implement these traits, or to implement them over the types directly?

I noticed that for `Task` there are some fields for which this might not be trivial, but for `TaskData` at least it appears that the derive macro should be able to handle it out-of-the-box, since it's basically just a `uuid::Uuid` + `HashMap<String, String>`.

I understand that the aim of the library is for direct manipulation of tasks via this storage layer, but this change would also enable users of the crate to use the same types for end-to-end communication. For example, being able to list pending tasks in a request to a backend, and then send a request with `Operations` already correctly encoded for it 🙂

miguelpduarte:
As per the suggestion in #624 I actually noticed that with `serde`'s `rc` feature we can get this quite easily.

I'm not 100% sure if this is in line with the crate's philosophy, but it seems to me that at least having the traits implemented for `TaskData` makes sense from a perspective of the library's ease of use.

Closes #624

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -42,6 +42,8 @@ cloud = []
 bundled = ["rusqlite/bundled"]
 # use native CA roots, instead of bundled
 tls-native-roots = ["ureq/native-certs"]
+# Expose serde::{Serialize, Deserialize} for crate types.
+serde = []
 
 [package.metadata.docs.rs]
 all-features = true
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -43,7 +43,10 @@ bundled = ["rusqlite/bundled"]
 # use native CA roots, instead of bundled
 tls-native-roots = ["ureq/native-certs"]
 # Expose serde::{Serialize, Deserialize} for crate types.
-serde = []
+# Note: Task's depmap will be copied to allow for freestanding deserialization.
+# This means that after Deserialization this data will not be de-duplicated.
+# See also: https://serde.rs/feature-flags.html#rc
+serde = ["serde/rc"]
 
 [package.metadata.docs.rs]
 all-features = true
--- a/src/depmap.rs
+++ b/src/depmap.rs
@@ -5,6 +5,7 @@ use uuid::Uuid;
 /// This information requires a scan of the working set to generate, so it is
 /// typically calculated once and re-used.
 #[derive(Debug, PartialEq, Eq, Clone)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
 pub struct DependencyMap {
     /// Edges of the dependency graph.  If (a, b) is in this array, then task a depends on task b.
     edges: Vec<(Uuid, Uuid)>,
--- a/src/task/data.rs
+++ b/src/task/data.rs
@@ -15,6 +15,7 @@ use uuid::Uuid;
 /// See the documentation for [`crate::Replica`] for background on the `ops` arguments to methods
 /// on this type.
 #[derive(Debug, PartialEq, Eq, Clone)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
 pub struct TaskData {
     uuid: Uuid,
     taskmap: TaskMap,
--- a/src/task/task.rs
+++ b/src/task/task.rs
@@ -26,6 +26,7 @@ use uuid::Uuid;
 /// See the documentation for [`crate::Replica`] for background on the `ops` arguments to methods
 /// on this type.
 #[derive(Debug, Clone)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
 pub struct Task {
     // The underlying task data.
     data: TaskData,


djmitche (changes requested):
Sorry for not replying for a while! I think allowing serialization is a good idea, full stop -- it should be easy for library users to get a task and then make that available to non-rust code. That really only makes sense for `TaskData`, though -- `Task` is more of an active type, with more "private" fields. The bits that can be effectively serialized are precisely those in `TaskData`. Serializling the depmap doesn't make sense for anyone consuming the serialized content, and certainly doesn't make sense to re-serialize as input from an external source.

As far as deserialization goes, I'm not sure there's much use for that -- a TaskData object is never passed _into_ a Replica. That said, it doesn't hurt to include that capability (aside from compile times!).

So I think I would prefer this PR with just the `src/task/data.rs` hunk, and not the other two.

You mentioned serializing Operations as well. I think that makes sense, and can imagine a case where someone builds a simple REST API around TaskChampion, exposing TaskData as output, and taking Operations as input. So, serializing and deserializing those makes good sense.

miguelpduarte:
> Sorry for not replying for a while! I think allowing serialization is a good idea, full stop -- it should be easy for library users to get a task and then make that available to non-rust code. That really only makes sense for `TaskData`, though -- `Task` is more of an active type, with more "private" fields. The bits that can be effectively serialized are precisely those in `TaskData`. Serializling the depmap doesn't make sense for anyone consuming the serialized content, and certainly doesn't make sense to re-serialize as input from an external source.
> As far as deserialization goes, I'm not sure there's much use for that -- a TaskData object is never passed _into_ a Replica. That said, it doesn't hurt to include that capability (aside from compile times!).
> 
> So I think I would prefer this PR with just the `src/task/data.rs` hunk, and not the other two.

At the time of creating the PR I was not yet very deeply familiar with the TaskChampion internals and thus just "blanket added" the trait derives for both, after noticing that they were "missing".

Now, although I do understand and agree about the conceptual usage of `Task`, I thought that it made sense from a "API consumer" perspective to be able to use the "richer type" -- As mentioned in the TaskChampion docs, `Task` has a nicer, higher-level API, including nicer getter and setter methods.

If the conclusion is that this still doesn't make sense re. TC's data model, then I can instead add a `impl From<TaskData> for Task` that just creates it with an empty `depmap`. From what I saw in the code, this does not cause "data loss" _per se_, as e.g. `Task::get_dependencies()` does not use the `depmap` and instead just accesses the underlying `TaskData` properties directly.

However, this would result in a "loss" of functionality for this barebones `Task`, as someone doing this would need to reimplement the synthetic tags functionality, which _feels_ redundant.

Maybe it makes sense to shine some light on the intention of these changes (and the overall proposal in #624):  
- I am trying my hand at building a mobile app to make using TaskWarrior on my phone less painful (using it in a terminal via Termux "works" - but the smaller width breaks a lot of the reports).  
Given that I have some Rust experience, I thought that it would make sense to build on top of TaskChampion directly, as then I could easily share the types all the way through my code.
- Due to the technology choices made ([Tauri](https://tauri.app/)), I have a sort of client<>server communication pattern in place (front-end and back-end) where it is extremely helpful to just be able to e.g. expose an endpoint for "list pending tasks" that returns a `Vec<Task>`, going over JSON, which then allows me to easily use all the existing methods on `Task` to display and manipulate data. I can then send these operations back to the "back-end" to ensure that everything is persisted and up-to-date 🙂
- So, the back-end in this situation would be using TaskChampion as intended (and as e.g. TaskWarrior uses it, directly), whereas the front-end is basically just using it to get "nicer" types and make the implementation smoother. (As a related note, I have some local changes that I'll post as a PR today to make `tokio` only required when it is actually needed, building on top of #620 -> This is necessary because when building Rust for WASM it is extremely difficult or even just outright impossible to carry `tokio` as a dependency 😕)
- (note: This is not open-source yet as the app isn't yet public. I'm trying to get it to a usable state for my own simple use-cases first, but I'm very inclined to do so once it reaches at least some initial level of maturity 😁)

> You mentioned serializing Operations as well. I think that makes sense, and can imagine a case where someone builds a simple REST API around TaskChampion, exposing TaskData as output, and taking Operations as input. So, serializing and deserializing those makes good sense.

I noticed in the meantime that those already have the required traits implemented for them -- so no changes would be required here. Which then leads me to think - do the proposed changes make sense as a separate `serde` feature after all?

They are adding trait implementations which are not _required_ for the internal functioning of TaskChampion itself, and are thus "nice to have"s, and using a non-`default` `feature` feels accurate.  
On the other hand, gating them behind a `feature` might cause confusion to anyone building on top of this crate, if they expect that to be available out-of-the-box.

---

@djmitche Sorry for the long response, but I do think that some more conversation might be useful, as I believe that these changes (plus the ones in the aforementioned upcoming PR) would help the "usability" of TaskChampion as a library, enabling use-cases such as mine: Building tools on top of this awesome software and tech!

In the meantime, I will split the proposed changes into 2 separate commits, one for each type. This way it will be easier to drop or cherry-pick them based on the decision made from this discussion 🙂

Thanks for the review and all the hard work on TaskChampion and TaskWarrior! It is very much appreciated - and I'm sure I'm not the only one who believes this 😄

djmitche:
No worries about the long response - I'm happy to have the context!

My concern with exposing all of `Task` in a readable format like JSON is that it forms a public API. Hyrum's law says that someone will come to depend on the details of that API (such as the format of the depmap). It also means we need to think about compatibility between the two sides of the API. If both sides are Rust, must those be the same version of TaskChampion? A the moment. `TaskData` is part of the [documented API](https://gothenburgbitfactory.org/taskchampion/tasks.html) of TaskChampion, so I'm pretty confident exposing that.

If I understand correctly, both the front and backends are Rust here, which is pretty cool! But, I don't think it's practical to make the existing `Task` type usable on both sides of a serialization API.

Instead, I think it would be best for this API to be private to your app, which controls both sides of the connection. Maybe that means serializing a task as TaskData plus whatever metadata is useful -- tags, maybe status as an enum, priority, etc. Then modifications could either be creation of Operations on the frontend, or API method calls to the backend that generate operations. So your fronend would basically have its own Task type that could have similar methods to the TaskChampion type.

Regarding use of depmap, `Task` uses it for `is_blocked` and `is_blocking`. Just leaving it blank would make those methods return incorrect information (and thus would change the synthetic tags - or maybe that's what you meant by "reimplement the synthetic tags functionality"?). Converting a TaskData to Task probably requires access to the Replica so that it can provide (and fetch if necessary) the depmap -- so, not quite an `impl<Into>`, but perhaps a `fn into_task(self, &mut Replica) -> Result<Task>`.

Regarding features, yes, TaskChampion already requires JSON support for the default features, so feature-gating this new functionality is not required. I should have noticed that earlier!

PR #627, opened 2025-10-12
fix: Scope tokio dependency to relevant features

miguelpduarte:
This allows e.g. depending on TaskChampion with `default_features = false` and being able to operate on the types directly.

This may be useful when e.g. consuming TaskChampion-shaped data from an API.

It also makes it possible to build `TaskChampion` for WASM targets, as long as [the notes for `uuid` are followed](https://docs.rs/uuid/latest/uuid/#webassembly).

(As per my comment in #626 🙂)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,15 +25,15 @@ sync = ["server-sync", "server-gcp", "server-aws", "server-local"]
 # Support for sync to a server
 server-sync = ["encryption", "dep:ureq", "dep:url"]
 # Support for sync to GCP
-server-gcp = ["cloud", "encryption", "dep:google-cloud-storage"]
+server-gcp = ["cloud", "encryption", "dep:google-cloud-storage", "dep:tokio"]
 # Support for sync to AWS
-server-aws = ["cloud", "encryption", "dep:aws-sdk-s3", "dep:aws-config", "dep:aws-credential-types"]
+server-aws = ["cloud", "encryption", "dep:aws-sdk-s3", "dep:aws-config", "dep:aws-credential-types", "dep:tokio"]
 # Suppport for sync to another SQLite database on the same machine
 server-local = ["storage-sqlite"]
 # Support for all task storage backends
 storage = ["storage-sqlite"]
 # Support for SQLite task storage
-storage-sqlite = ["dep:rusqlite"]
+storage-sqlite = ["dep:rusqlite", "dep:tokio"]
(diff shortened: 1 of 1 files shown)


miguelpduarte:
Hmm, not sure what to do about the failing tests since they all depend on `tokio`, but it will become annoying and bothersome to gate all of that behind `cfg` 😕

I had this building locally for the use-case mentioned in the other PR, but also now noticed that #617 exists - so it might make sense to "join forces" with the work done there, although there hasn't been activity in the PR for some time.

I opened this mostly to share the changes I had locally to make things work for me. If this doesn't make sense to pursue, I can close it or change it to a draft PR, as per maintainers' guidance 🙂

djmitche:
@geofflittle is also working on WASM support, such as in #617, so perhaps there's an opportunity to collaborate?

Regarding the tests, could `tokio` just be a dev-dependency?

miguelpduarte:
> @geofflittle is also working on WASM support, such as in #617, so perhaps there's an opportunity to collaborate?
> 
> Regarding the tests, could `tokio` just be a dev-dependency?

That's a good point and a very straightforward fix. I guess I missed it because I was just laser-focused on my front-end use-case. I'll update the PR with that soon-ish and then we can see where to take things if the tests start passing with no issues 🙂

djmitche:
(flag me for review again when this is ready!)

geofflittle:
Hi-o! Posted a PR that brings us closer to WASM-compatibility: #628. We need `Server` to be `async` so that we can write a browser-based impl to provide to `Replica`.

djmitche:
Hi @miguelpduarte -- how is it going? I got pretty busy for a bit and lost track of this work. Will you be able to finish it up?

djmitche:
I ended up landing something like this in #637, so I think this can be closed.

miguelpduarte:
> Hi @miguelpduarte -- how is it going? I got pretty busy for a bit and lost track of this work. Will you be able to finish it up?

@djmitche sorry for not replying earlier, a lot of other things came up, namely at $WORK and I was unfortunately unable to finish this in the meantime. Thanks for following up.

I think that despite #637 there might still be value in this for someone consuming "only the types" defined in TaskChampion, such as when defining or using an API.  

I believe we've already discussed this in another location here, so that might be the best place to continue the conversation instead 🙂 

Thanks for all the work in TaskChampion!

PR #628, opened 2025-10-13
refactor: Make `Server` async

geofflittle:
This PR refactors the `Server` trait and its implementations to be async.

**Key Changes**
* **Async Traits:** The core `Server` trait and the underlying `cloud::Service` trait are now defined with `async_trait`. All of their methods are now `async fn`.
* **Async Iterator Pattern:** To support asynchronous pagination, the standard `Iterator` pattern for listing objects has been replaced. A new `cloud::iter::AsyncObjectIterator` trait was introduced with an `async fn next()` method. The list methods on the `Service` and `Server` traits now return a `Box<dyn AsyncObjectIterator>`.
* **Removal of Blocking Calls:** The `AwsService` and `GcpService` implementations have been refactored to remove the embedded `tokio::runtime::Runtime` and all calls to `rt.block_on`. All cloud SDK operations now use `.await` directly for non-blocking I/O.
* **Implementation Updates:** `Server` implementations have been updated to be `async`, including `CloudServer` , `LocalServer` , `SyncServer` , and the `TestServer`.
* **ServerConfig:** The primary entry point, `ServerConfig::into_server`, is now an `async` function to support the async initialization of cloud clients.
* **Tests:** The test suite has been migrated to `#[tokio::test]`, with test functions and setup helpers (`make_server`) becoming `async`.

### DIFF
--- a/src/server/cloud/gcp.rs
+++ b/src/server/cloud/gcp.rs
@@ -1,16 +1,16 @@
 use super::service::{validate_object_name, ObjectInfo, Service};
 use crate::errors::Result;
+use crate::server::cloud::iter::AsyncObjectIterator;
+use async_trait::async_trait;
 use google_cloud_storage::client::google_cloud_auth::credentials::CredentialsFile;
 use google_cloud_storage::client::{Client, ClientConfig};
 use google_cloud_storage::http::error::ErrorResponse;
 use google_cloud_storage::http::Error as GcsError;
 use google_cloud_storage::http::{self, objects};
-use tokio::runtime::Runtime;
 
 /// A [`Service`] implementation based on the Google Cloud Storage service.
 pub(in crate::server) struct GcpService {
     client: Client,
-    rt: Runtime,
     bucket: String,
 }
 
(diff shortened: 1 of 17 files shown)


src/server/encryption.rs
```diff
@@ -59,7 +57,7 @@ impl Cryptor {
         } = payload;
 
         let mut nonce_buf = [0u8; aead::NONCE_LEN];
-        self.rng
+        rand::SystemRandom::new()
```
djmitche: This change seems fine but orthogonal to the purpose of the PR. Was it required for some reason?

It's fine to leave in -- I'm just curious.

src/server/cloud/gcp.rs
djmitche, line 54: It looks like `rt.block_on` wasn't removed in this file?

djmitche (changes requested):
This looks great - just a missed bit!

src/server/encryption.rs
geofflittle: I'm trying to remember, I think I had made this change in the original WASM attempt but I can't remember why. I'll remove it to make the diff smaller.

src/server/cloud/gcp.rs
geofflittle: Removing

geofflittle: Removed!

djmitche (approved):
Awesome, thanks!!

PR #634, opened 2025-11-17
Test tls-native-roots and fix it for clouds

### LINKED ISSUE #608: No CA certificates found panic when setting up AWS S3 sync
Hey, I am trying to add AWS S3 syncing integration to [taskchamp](https://github.com/marriagav/taskchamp), note that this has the taskchampion rust binary running in iOS via this [bridge](https://github.com/marriagav/task-champion-swift/blob/main/taskchampion-swift/src/lib.rs)

However, I keep getting the following panic when setting up the sync:

```
thread '<unnamed>' panicked at /Users/<user>/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/hyper-rustls-0.24.2/src/config.rs:48:9:
no CA certificates found
stack backtrace:
0: \_rust_begin_unwind
1: core::panicking::panic_fmt
2: <rustls::builder::ConfigBuilder<rustls::client::client_conn::ClientConfig,rustls::builder::WantsVerifier> as hyper_rustls::config::ConfigBuilderExt>::with_native_roots
3: aws_smithy_runtime::client::http::hyper_014::default_connector::default_tls
4: once_cell::imp::OnceCell<T>::initialize::{{closure}}
5: once_cell::imp::initialize_or_wait
6: once_cell::imp::OnceCell<T>::initialize
7: aws_smithy_runtime::client::http::hyper_014::default_connector::https
8: <aws_smithy_runtime::client::http::hyper_014::HyperClient<F> as aws_smithy_runtime_api::client::http::HttpClient>::validate_base_client_config
9: aws_smithy_runtime_api::client::runtime_components::RuntimeComponentsBuilder::validate_base_client_config
10: aws_sdk_s3::client::Client::from_conf
11: aws_sdk_s3::client::Client::new
12: taskchampion::server::cloud::aws::AwsService::new
13: taskchampion::server::config::ServerConfig::into_server
14: taskchampion_swift::Replica::sync_aws
```

It looks to me from the stacktrace that the way its setup AWS is using native TLS root certs (which it cannot find in iOS) instead of the bundled certs. Note that I did not enable the `tls-native-roots` feature introduced in https://github.com/GothenburgBitFactory/taskchampion/issues/460.

I dug a bit into it and think that we might need to override the HttpClient passed to aws config here: https://github.com/awslabs/aws-sdk-rust/blob/cccaf7ce452b84ba3800837c63280aec531c67a5/sdk/aws-config/src/lib.rs#L398

Appreciate the help : )

djmitche:
This adds tests to ensure TC is correctly implementing tls-native-roots and tls-webpki-roots for all server backends, then fixes it for AWS and GCP.

To save on binary size, it introduces a new feature, `tls-webpki-roots`, which when disabled (via `--no-default-features`) will omit the webpki dependency and thus save the binary size of bundling all those certs. This constitutes a breaking change for users building with `--no-default-features`.

Fixes #608.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,16 +18,16 @@ members = [ "xtask" ]
 resolver = "2"
 
 [features]
-default = ["sync", "bundled", "storage"]
+default = ["sync", "bundled", "storage", "tls-webpki-roots"]
 
 # Support for all sync solutions
 sync = ["server-sync", "server-gcp", "server-aws", "server-local"]
 # Support for sync to a server
 server-sync = ["encryption", "dep:ureq", "dep:url"]
 # Support for sync to GCP
-server-gcp = ["cloud", "encryption", "dep:google-cloud-storage"]
+server-gcp = ["cloud", "encryption", "dep:google-cloud-storage", "dep:reqwest-middleware"]
 # Support for sync to AWS
-server-aws = ["cloud", "encryption", "dep:aws-sdk-s3", "dep:aws-config", "dep:aws-credential-types"]
+server-aws = ["cloud", "encryption", "dep:aws-sdk-s3", "dep:aws-config", "dep:aws-credential-types", "dep:aws-smithy-runtime"]
 # Suppport for sync to another SQLite database on the same machine
 server-local = ["storage-sqlite"]
 # Support for all task storage backends
@@ -40,8 +40,10 @@ encryption = ["dep:ring"]
 cloud = []
 # static bundling of dependencies
 bundled = ["rusqlite/bundled"]
-# use native CA roots, instead of bundled
-tls-native-roots = ["ureq/native-certs"]
+# use CA roots built into the library for all HTTPS access
+tls-webpki-roots = ["dep:hyper-rustls", "hyper-rustls/webpki-roots", "dep:reqwest", "reqwest/rustls-tls-webpki-roots"]
+# use native CA roots (from the filesystem), instead of bundled for all HTTPS access
+tls-native-roots = ["ureq/native-certs", "dep:hyper-rustls", "hyper-rustls/rustls-native-certs", "dep:reqwest", "reqwest/rustls-tls-native-roots"]
 
 [package.metadata.docs.rs]
 all-features = true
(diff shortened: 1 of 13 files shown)


djmitche:
I didn't really mean to click on Copilot to review this but let's see what it comes up with.

This is all a bit of a mess in terms of dependencies. I filed #635 to try to reduce this. Ideally we'd have one and only one copy of hyper-rustls, for example, and no webpki when that's not enabled.

src/errors.rs
```diff
@@ -56,6 +56,7 @@ other_error!(google_cloud_storage::client::google_cloud_auth::error::Error);
 other_error!(aws_sdk_s3::Error);
 #[cfg(feature = "server-aws")]
 other_error!(aws_sdk_s3::primitives::ByteStreamError);
```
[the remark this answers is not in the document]
djmitche: It doesn't fail to compile, because reqwest is always included (due to how the features are defined -- there's no way to say "depend on reqwest only if both `tls-*-roots` and `server-gcp` are enabled"). However, we can avoid generating this impl when it's not needed, which should strip some code when `server-gcp` is not enabled.

djmitche:
That was useful!

djmitche:
OK, checks are green now. The "Expected" check has changed names, so I'll update that when the PR lands.

Cargo.toml
```diff
@@ -40,8 +40,10 @@ encryption = ["dep:ring"]
 cloud = []
 # static bundling of dependencies
 bundled = ["rusqlite/bundled"]
-# use native CA roots, instead of bundled
-tls-native-roots = ["ureq/native-certs"]
+# use CA roots built into the library for all HTTPS access
+tls-webpki-roots = ["dep:hyper-rustls", "hyper-rustls/webpki-roots", "dep:reqwest", "reqwest/rustls-tls-webpki-roots"]
```
ryneeverett, line 44: To me, calling this "webpki" in contrast to "native" seems odd -- it's "web public key infrastructure" either way, right? -- and something like "tls-vendored-roots" or "tls-bundled-roots" would make more sense.

src/server/cloud/gcp.rs
```diff
@@ -30,13 +35,31 @@ impl GcpService {
         bucket: String,
         credential_path: Option<String>,
     ) -> Result<Self> {
-        let config: ClientConfig = if let Some(credentials) = credential_path {
+        #![allow(unused)]
+        let mut config = ClientConfig::default();
+        #[cfg(feature = "tls-native-roots")]
+        let (native, webpki) = (true, false);
+        #[cfg(not(feature = "tls-native-roots"))]
+        let (native, webpki) = (false, true);
```
ryneeverett: I don't follow the point of these booleans. Aren't `native` and `webpki` always `true` when they're conditionally compiled?

ryneeverett (approved):

Cargo.toml
djmitche: I'm mirroring what other crates, including reqwest and hyper, call it. It refers to https://docs.rs/webpki-root-certs/1.0.4/webpki_root_certs/. I think the consistency of naming is marginally more important than clarity. Also, we already have a "bundled" feature which bundles sqlite3, so that name might be confusing.

src/server/cloud/gcp.rs
djmitche: Yes! I missed this cleanup when I added the `tls-webpki-roots` feature.

PR #637, opened 2025-11-30
Add a WASM build to CI

djmitche:
This does not support a full set of features yet (only sqlite storage, which will not be much use in WASM, and local sync), but more functionality is coming.

r? @geofflittle particularly for the way I've solved the uuid randomness dependency.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -68,12 +68,15 @@ serde_json = "^1.0"
 serde = { version = "^1.0.147", features = ["derive"] }
 strum = "0.27"
 strum_macros = "0.27"
-tokio = { version = "1", features = ["rt-multi-thread", "macros", "sync"] }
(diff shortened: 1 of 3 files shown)


Cargo.toml
```diff
@@ -68,16 +68,20 @@ serde_json = "^1.0"
...
+[target.'cfg(target_arch = "wasm32")'.dependencies]
+uuid = { version = "*", features = ["js"] }
+
 [dev-dependencies]
 proptest = "^1.9.0"
 tempfile = "3"
 rstest = "0.26"
 pretty_assertions = "1"
 libc = "*"
+tokio = { version = "*", features = ["rt"] }
```
geofflittle: Should we pin this to a version for the same reason? If `tokio` publishes a new major version with a breaking change?

PR #638, opened 2025-12-02
Fix warnings with --no-default-features

djmitche:
The warning in config.rs occurs when there are no server features enabled, in which case any attempt to construct `ServerConfig` will fail.

The errors.rs warning is just an unused import since the test that uses it is conditional.

These warnings were annoying me while working on WASM support!

### DIFF
--- a/src/errors.rs
+++ b/src/errors.rs
@@ -82,6 +82,7 @@ pub(crate) type Result<T> = std::result::Result<T, Error>;
 
 #[cfg(test)]
 mod test {
+    #[cfg(feature = "server-sync")]
     use super::*;
 
     #[cfg(feature = "server-sync")]
--- a/src/server/config.rs
+++ b/src/server/config.rs
@@ -97,6 +97,8 @@ pub enum ServerConfig {
 impl ServerConfig {
     /// Get a server based on this configuration
     pub async fn into_server(self) -> Result<Box<dyn Server>> {
+        // This expression is unreachable if no server features are enabled.
+        #[allow(unreachable_code)]
         Ok(match self {
             #[cfg(feature = "server-local")]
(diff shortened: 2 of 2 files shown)


ryneeverett (commented):
I don't really understand what code is considered unreachable in config.rs, but would adding a default case that throws an error shut up the warning and perhaps be more expressive?

djmitche:
AIUI when there are no variants of the ServerConfig enum, it is an uninhabited type, so any expression containing a value of that type is unreachable. I don't think a dummy match arm would help, as it is the match statement itself that is unreachable. The error is
```
warning: unreachable call
   --> src/server/config.rs:100:9
    |
100 |           Ok(match self {
    |  _________^^_-
    | |         |
    | |         unreachable call
101 | |             #[cfg(feature = "server-local")]
102 | |             ServerConfig::Local { server_dir } => Box::new(LocalServer::new(server_dir)?),
103 | |             #[cfg(feature = "server-sync")]
...   |
133 | |             ),
134 | |         })
    | |_________- any code following this expression is unreachable
    |
    = note: `#[warn(unreachable_code)]` (part of `#[warn(unused)]`) on by default
```

And this is as we want it: if all server-* features are disabled, library users cannot call `ServerConfig::Something{..}.into_server()`

Another option to express this would be
```diff
impl ServerConfig {
+    #[cfg(any(
+        feature = "server-local",
+        feature = "server-gcp",
+        feature = "server-aws",
+        feature = "server-sync"
+    ))]
     /// Get a server based on this configuration
     pub async fn into_server(self) -> Result<Box<dyn Server>> {
-        #[allow(unreachable_code)]
         Ok(match self {
```
but this leaves some imports unused, and adding that long `#[cfg(..)]` conditional for the `use` statements starts to get very verbose.

I'll add a comment to the `#[allow(..)]`

ryneeverett (approved):

Issue #635, opened 2025-11-17
Reduce the number of HTTP clients


djmitche:
We use HTTP for server-sync, server-aws, and server-gcp. They all use different HTTP clients! Let's try as much as possible to limit that to one client.

djmitche:
I noticed that the `*` specification for hyper_rustls doesn't make sense now that we use that crate directly in the source, and causes diamond-dependency problems when running `cargo update`. Pinning it to `0.24.2` fixes this temporarily, but that should be addressed more throroughly.

PR #639, opened 2025-12-02
Pin the version of hyper-rustls

djmitche:
See #635. This is not a long-term fix.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -58,7 +58,7 @@ byteorder = "1.5"
 chrono = { version = "^0.4.38", features = ["serde"] }
 flate2 = "1"
 google-cloud-storage = { version = "0.24.0", default-features = false, features = ["rustls-tls", "auth"], optional = true }
-hyper-rustls = { version = "*", optional = true } # AWS SDK selects a version
+hyper-rustls = { version = "0.24.2", optional = true }
 log = "^0.4.17"
 reqwest = { version = "*", optional = true } # GCP SDK selects a version
 reqwest-middleware = { version = "*", optional = true } # GCP SDK selects a version


Cargo.toml
```diff
@@ -58,7 +58,7 @@ byteorder = "1.5"
 chrono = { version = "^0.4.38", features = ["serde"] }
 flate2 = "1"
 google-cloud-storage = { version = "0.24.0", default-features = false, features = ["rustls-tls", "auth"], optional = true }
-hyper-rustls = { version = "*", optional = true } # AWS SDK selects a version
+hyper-rustls = { version = "0.24.2", optional = true } # AWS SDK selects a version
```
ryneeverett: "AWS SDK selects a version" <-- no longer true, right?

ryneeverett (commented):
I don't understand how pinning a version -- an additional constraint -- improves a dependency hell situation.

djmitche:
Good point about that no longer being true!

I'm a little unclear on how this works, too. Indirectly, there are dependencies on two semver-incompatible versions of this crate, so Rust builds them both and considers their types distinct. I think the `*` meant that when the `taskchampion` crate referred to `hyper_rustls` (which it does to construct a client for AWS), it got the more recent version, while AWS was expecting the older version.

This obviously works in `main`, but I ran into this issue when building for WASM, and couldn't quite pin down what change in the WASM patch caused the dependency to shift. But, this cleared it up so I considered it a break-fix for #635.

ryneeverett:
So ideally we'd want to say "cargo, give me the same version of hyper_rustls that my aws_sdk_s3 dependency is using", right?

ryneeverett (approved):

djmitche:
Yes, in fact in #635 I'm hoping we can use the same HTTP library for all three uses of HTTP (AWS, GCP, and Sync).

PR #642, opened 2025-12-06
Factor out the Send storage wrapper

djmitche:
In #620, @geofflittle created a fancy wrapper for the `!Send` SQLite libraries to allow them to be used in more common `Send` runtimes (where e.g., `spawn(fut)` expects a `Send` future). The SQLite C library uses thread-local storage, and thus must always run in the same thread, which in Rust means it's `!Send`. Similarly, a planned IndexedDB backend for use in WASM builds contains raw pointers and thus is also `!Send`.

This PR factors the storage wrapper out so that it can be used to support other storage backends (in particular, a planned IndexedDB backend for use in WASM builds). It goes on to support async methods on the wrapped backend, and an async constructor for that backend. It also makes some small tweaks: removing the response channel from the Rollback message, and wrapping the `is_empty` method.

There is one breaking change: `SqliteStorage::new` is now async, as it must wait for the initialization to take place in the thread.

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -29,7 +29,7 @@ jobs:
 
       - uses: actions-rs/toolchain@v1
         with:
-          toolchain: "1.82.0" # MSRV
+          toolchain: "1.85.0" # MSRV
           override: true
           components: clippy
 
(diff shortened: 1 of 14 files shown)


geofflittle:
I'm sorry I was sitting on this--here's an alternative impl for making the actor pattern generic: #643. It avoids the trait dupe.

ryneeverett (approved):

PR #643, opened 2025-12-06
feat: Refactor `SqliteStorage` to use generic actor abstraction

geofflittle:
This PR generalizes the "Actor Model" pattern—previously hardcoded in `SqliteStorage`—into a reusable `ActorStorage<A, T>` struct. This allows `!Send` storage backends (like the upcoming `IndexedDB` implementation) to satisfy the `Storage: Send` trait requirement via message passing, which is a prerequisite for WASM support.

**Key Changes**
* **New `src/storage/actor.rs`:** Introduced `ActorStorage` (a generic, `Send`-safe wrapper) and `ActorTxn` (a transaction proxy).
* **Refactored SQLite:** Updated `SqliteStorage` to use this generic wrapper instead of custom actor logic.
* **New Traits:** Defined `ActorMessage` and `TxnMessage` traits to allow backends to define their own message types.

### DIFF
--- a/src/errors.rs
+++ b/src/errors.rs
@@ -1,5 +1,5 @@
 #[cfg(feature = "storage-sqlite")]
-use crate::storage::sqlite::ActorMessage;
+use crate::storage::sqlite::SqliteActorMessage;
 use std::io;
 use thiserror::Error;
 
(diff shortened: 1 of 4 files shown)


djmitche (commented):
Wow, funny that we both worked on basically the same task.

I'd be happy to merge your version, but there are a few things I like about my version:
 - It abstracts away the "actor" bit of things so the externally-visible difference is just that the Wrapped* traits are `(?Send)`. There's no need for the wrapped types to know about or implement messages, channel dispatch, etc.
 - You mention not duplicating a trait in my PR, but in this patch `ActorMessage` is essentially a duplicate of the `StorageTxn` trait, just with a different type. So, I'm not sure that advantage is real.
 - For IndexedDb, I think we will want an async constructor; `ActorStorage::new_internal` does not support this.

Are there advantages to your approach I might adopt in my PR?

geofflittle:
Great minds! On a deeper look, your solution is better--there's nothing you should adopt from my PR.

PR #645, opened 2025-12-10
Factor server implementations to use a single HTTP library

### LINKED ISSUE #635: Reduce the number of HTTP clients

djmitche:
This refactors all three remote server implementations (AWS, GCP, Server) to use reqwest. This reduces the number of dependencies and simplifies configuration of HTTP parameters like TLS certs. It also makes the `server-sync` implementation actually async, replacing the blocking use of ureq.

Mostly, this involves some adaptation to make the cloud SDKs use a provided HTTP client, but this is relatively straightforward. The Server implementation sees a more substantial overhaul, and in the process gets tests to ensure it's behaving reasonably.

A simple binary built against the library was 25MB before and 18MB after this change.


djmitche:
This includes the MSRV bump (#646) so feel free to review only the second commit.

src/errors.rs
```diff
@@ -50,51 +50,28 @@ other_error!(google_cloud_storage::client::google_cloud_auth::error::Error);
...
-        let err = ureq::Error::Status(
-            418,
-            ureq::Response::new(418, "I Am a Teapot", "uhoh").unwrap(),
-        );
-        assert_eq!(
-            Error::from(err).to_string(),
-            "Server Error: https://example.com/ responded with 418 I Am a Teapot"
-        );
-    }
-}
```
ryneeverett, line 100: Seems like we could be testing the reqwest_error_status here.

ryneeverett (approved):

src/errors.rs
djmitche: I couldn't find any straightforward way to create a new reqwest::Error, besides doing something that would involve network access, and that seems like it might be flaky.

PR #646, opened 2025-12-10
Bump the MSRV to 1.88.0

djmitche:
This is from June 26, 2025 which is pretty recent! However, newer versions of `aws-config` require it, and those support work on #635, so I think it's worthwhile.


ryneeverett (approved):
I like the new string formatting!

djmitche:
> I like the new string formatting!

So does Clippy :)

PR #649, opened 2025-12-15
Add an IndexedDB storage implementation

djmitche:
This is tested in both browsers in CI using `wasm-pack test`, and passes all existing storage tests.

Because proptest is not compatible with WASM, this modifies some existing tests to ensure they only build on non-WASM platforms.

Refs #599.

### DIFF
--- a/src/errors.rs
+++ b/src/errors.rs
@@ -57,6 +57,22 @@ impl<T: Sync + Send + 'static> From<tokio::sync::mpsc::error::SendError<T>> for 
     }
 }
 
+/// Convert [`idb::Error`] into [`Error::Database`]
+#[cfg(all(target_arch = "wasm32", feature = "storage-indexeddb"))]
+impl From<idb::Error> for Error {
+    fn from(err: idb::Error) -> Self {
+        Error::Database(err.to_string())
+    }
+}
+
+/// Convert [`serde_wasm_bindgen::Error`] into [`Error::Database`]
+#[cfg(all(target_arch = "wasm32", feature = "storage-indexeddb"))]
+impl From<serde_wasm_bindgen::Error> for Error {
+    fn from(err: serde_wasm_bindgen::Error) -> Self {
+        Error::Database(err.to_string())
+    }
+}
+
 /// Convert reqwest errors more carefully
 #[cfg(feature = "http")]
 impl From<reqwest::Error> for Error {
--- a/src/storage/indexeddb/schema.rs
+++ b/src/storage/indexeddb/schema.rs
@@ -1,0 +1,71 @@
+use crate::errors::Result;
+use idb::{Database, KeyPath};
+
+/// The IndexedDB database version. This can be incremented to trigger calls to
+/// `upgrade`.
+pub(super) const DB_VERSION: u32 = 1;
+
+pub(super) const TASKS: &str = "tasks";
+pub(super) const OPERATIONS: &str = "operations";
+pub(super) const OPERATIONS_BY_UUID: &str = "operations_by_uuid";
+pub(super) const OPERATIONS_BY_UNSYNCED: &str = "operations_by_unsynced";
+pub(super) const SYNC_META: &str = "sync_meta";
+pub(super) const WORKING_SET: &str = "working_set";
+
+/* Current Schema:
+ *
+ *  - Object store `tasks` stores TaskMaps as values, in the form of a JS object. Keys are UUIDs in
+ *    the form of a string.
+ *
+ *  - Object store `operations` stores values {uuid, operation, unsynced}, with `unsynced` omitted
+ *    for synced operations. Keys are auto-incrementing integers. The `uuid` and `unsynced`
+ *    properties are indexed.
+ *
+ *  - Object store `sync_meta` stores string values with string keys. At the moment it only stores
+ *    the base revision.
+ *
+ *  - Object store `working_set` stores UUIDs as strings, keyed by the working set ID as an integer.
+ */
+
+fn upgrade_0_to_1(db: Database) -> Result<()> {
+    // Create the `tasks` table
+    let mut params = idb::ObjectStoreParams::new();
+    params.key_path(None);
+    params.auto_increment(false);
+    db.create_object_store(TASKS, params)?;
+
+    // Create the `operations` table
+    let mut params = idb::ObjectStoreParams::new();
+    params.key_path(None);
+    params.auto_increment(true);
+    let ops = db.create_object_store(OPERATIONS, params)?;
+
+    ops.create_index(OPERATIONS_BY_UUID, KeyPath::new_single("uuid"), None)?;
+    ops.create_index(
+        OPERATIONS_BY_UNSYNCED,
+        KeyPath::new_single("unsynced"),
+        None,
+    )?;
+
+    // Create the `sync_meta` table
+    let mut params = idb::ObjectStoreParams::new();
+    params.key_path(None);
+    params.auto_increment(false);
+    db.create_object_store(SYNC_META, params)?;
+
+    // Create the `working_set` table
+    let mut params = idb::ObjectStoreParams::new();
+    params.key_path(None);
+    params.auto_increment(false);
+    db.create_object_store(WORKING_SET, params)?;
+
+    Ok(())
+}
+
+pub(super) fn upgrade(db: Database, old_version: u32) -> Result<()> {
+    if old_version == 0 {
+        upgrade_0_to_1(db)
+    } else {
+        unreachable!("old_version should be less than DB_VERSION");
+    }
+}
--- a/src/storage/indexeddb/storage.rs
+++ b/src/storage/indexeddb/storage.rs
@@ -1,0 +1,467 @@
+use super::schema;
+use crate::errors::{Error, Result};
+use crate::operation::Operation;
+use crate::storage::send_wrapper::{WrappedStorage, WrappedStorageTxn, Wrapper};
+use crate::storage::{Storage, StorageTxn, TaskMap, VersionId, DEFAULT_BASE_VERSION};
+use async_trait::async_trait;
+use idb::{CursorDirection, DatabaseEvent, Query, Transaction, TransactionMode};
+use serde::{Deserialize, Serialize};
+use std::collections::{HashMap, HashSet};
+use uuid::Uuid;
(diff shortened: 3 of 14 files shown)


src/storage/indexeddb/storage.rs
```diff
@@ -0,0 +1,439 @@
...
+    async fn get_next_working_set_number(&self) -> Result<usize> {
+        let working_set = self.idb_txn()?.object_store(schema::WORKING_SET)?;
+        let mut max: usize = 0;
+        let mut maybe_cursor = working_set.open_key_cursor(None, None)?.await?;
+        while let Some(cursor) = maybe_cursor {
+            let i = cursor.key()?.as_f64().ok_or_else(invalid)? as usize;
+            if i > max {
+                max = i;
+            }
+            maybe_cursor = cursor.next(None)?.await?;
```
geofflittle, line 135: Instead of iter'ing through the whole db, should `current_max_id` be denormalized into a separate obj in the db that's updated whenever items are added?

djmitche: Let's call that a followup? It would still require a table scan when removing an item from the working set, so I'm not sure it would be a net savings in performance.

src/errors.rs
```diff
@@ -57,6 +57,22 @@ impl<T: Sync + Send + 'static> From<tokio::sync::mpsc::error::SendError<T>> for
...
+    fn from(err: idb::Error) -> Self {
+        Error::Database(err.to_string())
+    }
+}
+
+/// Convert [`serde_wasm_bindgen::Error`] into [`Error::Other`]
+#[cfg(all(target_arch = "wasm32", feature = "storage-indexeddb"))]
+impl From<serde_wasm_bindgen::Error> for Error {
+    fn from(err: serde_wasm_bindgen::Error) -> Self {
+        Error::Database(err.to_string())
```
ryneeverett, line 72: The doc-comment says it's converted into `Error::Other`.

src/storage/indexeddb/schema.rs
```diff
@@ -0,0 +1,74 @@
...
+/* Current Schema:
+ *  - tasks
+ *    - values: taskmap
+ *    - key: string (uuid)
+ *  - operations
+ *    - values: js object containing {uuid, operation, unsynced} (with unsynced omitted for synced
+*    operations)
+ *    - key: int (auto increment)
+ *    - index: uuid
+ *    - index: unsynced
```
ryneeverett: Accidental second `index` field?

src/storage/indexeddb/storage.rs
```diff
@@ -0,0 +1,439 @@
...
+fn already_committed() -> Error {
+    Error::Database("transaction is already committed".into())
+}
+
+/// Return a new error indicating the operation was not found.
+fn operation_not_found() -> Error {
+    Error::Database("operation not found".into())
+}
+
+/// Return a new error indicating the operation was not found.
```
ryneeverett:
```suggestion
/// Return a new error indicating the operation was synced.
```

src/storage/indexeddb/storage.rs
```diff
@@ -0,0 +1,439 @@
...
+
+    // IndexedDB does not support indexing by a boolean value, so
+    // we store a number here instead. Since we want to index un-synced
+    // operations, this field is only present on such entries.
+    #[serde(skip_serializing_if = "unsynced_is_zero")]
+    #[serde(default)]
+    unsynced: u8,
+}
+
+/// Convert a JsValue to a Task.
```
ryneeverett:
```suggestion
/// Convert a JsValue to a Operation.
```

src/storage/indexeddb/storage.rs
```diff
@@ -0,0 +1,439 @@
...
+fn js2op(js: JsValue) -> Result<Operation> {
+    Ok(serde_wasm_bindgen::from_value::<StoredOperation>(js)?.operation)
+}
+
+/// Convert a Task to a JsValue.
```
ryneeverett:
```suggestion
/// Convert an Operation to a JsValue.
```

src/storage/indexeddb/storage.rs
```diff
@@ -0,0 +1,439 @@
...
+        let tasks = self.idb_txn()?.object_store(schema::TASKS)?;
+        if let Some(task) = tasks.get(Query::Key(uuid2js(uuid)?))?.await? {
+            Ok(Some(js2task(task)?))
+        } else {
+            Ok(None)
+        }
+    }
+
+    async fn get_pending_tasks(&mut self) -> Result<Vec<(Uuid, TaskMap)>> {
+        todo!()
```
ryneeverett: I didn't know about this macro. Handy! Did you mean to commit it though?

src/storage/indexeddb/storage.rs
```diff
@@ -0,0 +1,439 @@
...
+    async fn create_task(&mut self, uuid: Uuid) -> Result<bool> {
+        let tasks = self.idb_txn()?.object_store(schema::TASKS)?;
+        match tasks
+            .add(&task2js(HashMap::new())?, Some(&uuid2js(uuid)?))?
+            .await
+        {
+            Ok(_) => Ok(true),
+            // This assumes that all interesting errors other than ConstraintError are of other
+            // variants of idb::Error.
+            Err(idb::Error::DomException(_)) => Ok(false),
```
ryneeverett: Did you mean to match ConstraintError here? I'm still not sure I'd understand the comment...

src/storage/indexeddb/schema.rs
```diff
@@ -0,0 +1,74 @@
...
+use idb::{Database, KeyPath};
+
+/// The IndexedDB database version. This can be incremented to trigger calls to
+/// `upgrade`.
+pub(super) const DB_VERSION: u32 = 1;
+
+pub(super) const TASKS: &str = "tasks";
+pub(super) const OPERATIONS: &str = "operations";
+pub(super) const OPERATIONS_BY_UUID: &str = "operations_by_uuid";
+pub(super) const OPERATIONS_BY_UNSYNCED: &str = "operations_by_synced";
```
ryneeverett:
```suggestion
pub(super) const OPERATIONS_BY_UNSYNCED: &str = "operations_by_unsynced";
```

src/storage/indexeddb/schema.rs
ryneeverett:
```suggestion
 *    - unsynced
```

src/storage/indexeddb/storage.rs
```diff
@@ -0,0 +1,439 @@
...
+        let found_op = serde_wasm_bindgen::from_value::<StoredOperation>(cursor.value()?)?;
+        if found_op.unsynced == 0 {
+            return Err(operation_synced());
+        }
+        if found_op.operation != op {
+            return Err(operation_not_found());
+        }
+        cursor.delete()?.await?;
+        Ok(())
+    }
```
ryneeverett, line 315: Am I right to understand that this method can only remove the "last" operation?

src/storage/indexeddb/schema.rs
```diff
@@ -0,0 +1,74 @@
...
+ *    - key: int (auto increment)
+ *    - index: uuid
+ *    - index: unsynced
+ *  - sync_meta
+ *    - values: string
+ *    - keys: string
+ *  - working_set
+ *    - data: string (uuid)
+ *    - keys: int (id)
+ */
```
ryneeverett, line 28: More consistency on the singular/plural key(s) value(s) would be nice.

src/storage/indexeddb/schema.rs
```diff
@@ -0,0 +1,74 @@
...
+ *    - values: js object containing {uuid, operation, unsynced} (with unsynced omitted for synced
+*    operations)
+ *    - key: int (auto increment)
+ *    - index: uuid
+ *    - index: unsynced
+ *  - sync_meta
+ *    - values: string
+ *    - keys: string
+ *  - working_set
+ *    - data: string (uuid)
```
ryneeverett:
```suggestion
 *    - values: string (uuid)
```

src/storage/indexeddb/storage.rs
```diff
@@ -0,0 +1,439 @@
...
+    async fn add_to_working_set(&mut self, uuid: Uuid) -> Result<usize> {
+        let next_working_set_id = self.get_next_working_set_number().await?;
+        let working_set = self.idb_txn()?.object_store(schema::WORKING_SET)?;
+        working_set
+            .add(&uuid2js(uuid)?, Some(&next_working_set_id.into()))?
+            .await?;
+        Ok(next_working_set_id)
+    }
+
+    async fn set_working_set_item(&mut self, index: usize, uuid: Option<Uuid>) -> Result<()> {
```
ryneeverett: While the taskwarrior id is technically an index, I think this helps distinguish it from all the other "index" stuff going on here.

```suggestion
    async fn set_working_set_item(&mut self, id: usize, uuid: Option<Uuid>) -> Result<()> {
```

src/storage/indexeddb/storage.rs
```diff
@@ -0,0 +1,439 @@
...
+    fn drop(&mut self) {
+        if let Some(txn) = self.0.take() {
+            // Make an attempt to abort the transaction, but without any recourse
+            // if it fails.
+            let _ = txn.abort();
+        }
+    }
+}
+
+/// IndexedDbStorage .. TODO doc
```
ryneeverett: TODO

ryneeverett (approved):

src/storage/indexeddb/storage.rs
djmitche: Oh! There are no tests for that method! Yikes!

src/storage/indexeddb/storage.rs
djmitche: In the process of trying to explain this, I figured out the issue and can precisely match the expected error now.

src/storage/indexeddb/storage.rs
djmitche: Yes, per the [StorageTxn docs](https://docs.rs/taskchampion/latest/taskchampion/storage/trait.StorageTxn.html#tymethod.remove_operation)

src/storage/indexeddb/schema.rs
djmitche: I rewrote it to a clearer prose style.

PR #653, opened 2025-12-21
Add server-sync support on WASM

djmitche:
This supports server-sync with few changes by virtue of reqwest's WASM support.

The `wasm-pack test` machinery does not support starting a test server reachable from the browser, so there is no practical way to test this functionality on this platform. Since the implementation differs only slightly from that for non-web platforms, this is acceptable.

I experimented with supporting the cloud sync backends as well. I was able to get AWS working, at the cost of some shenanigans around Send and non-Send futures. However, GCP appears impossible. The current `google_cloud_storage` version we are using unconditionally requires `tokio` and its I/O features, which means `mio` which does not build on WASM. The newer and completely rewritten (see note in [the docs](https://docs.rs/google-cloud-storage/latest/google_cloud_storage/)) version of `google_cloud_storage` does not allow supplying an HTTP client, and thus also requires `mio`. Even if we could build these libraries for WASM, they would likely be huge and impractical to download and run in a web browser.

I think we should ship TaskChampion 3.0.0 without WASM cloud support, and then think of alternatives to support this. The binary size of the cloud libraries is also a problem on regular POSIX and Windows systems, so perhaps some kind of external interface would be best.

### DIFF
--- a/src/server/encryption.rs
+++ b/src/server/encryption.rs
@@ -27,6 +27,7 @@ impl Cryptor {
     }
 
     /// Generate a suitable random salt.
+    #[cfg(any(test, feature = "cloud"))] // server-sync uses the clientId as the salt.
     pub(super) fn gen_salt() -> Result<Vec<u8>> {
         let rng = rand::SystemRandom::new();
         let mut salt = [0u8; 16];
--- a/src/server/http.rs
+++ b/src/server/http.rs
@@ -3,11 +3,12 @@
 //! This contains some utilities to make using `reqwest` easier, including getting
 //! the correct TLS certificate store.
 
-use std::time::Duration;
-
 use crate::errors::Result;
 
-#[cfg(not(any(feature = "tls-native-roots", feature = "tls-webpki-roots")))]
+#[cfg(all(
+    not(target_arch = "wasm32"),
+    not(any(feature = "tls-native-roots", feature = "tls-webpki-roots"))
+))]
 compile_error!(
     "Either feature \"tls-native-roots\" or \"tls-webpki-roots\" must be enabled for HTTP client support."
 );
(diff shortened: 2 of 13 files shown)


src/server/encryption.rs
```diff
@@ -27,6 +27,7 @@ impl Cryptor {
     }
 
     /// Generate a suitable random salt.
+    #[cfg(any(test, feature = "cloud"))] // server-sync uses the clientId as the salt.
```
ryneeverett, line 30: Does this actually significantly reduce the build size or is this more of a best practice to always conditionally compile when possible? I'm also curious why you wouldn't use `not(feature = "server-sync")` which would seem to express the intent better.

src/server/http.rs
```diff
@@ -3,19 +3,22 @@
...
 compile_error!(
     "Either feature \"tls-native-roots\" or \"tls-webpki-roots\" must be enabled for HTTP client support."
 );
 
 static USER_AGENT: &str = concat!(env!("CARGO_PKG_NAME"), "/", env!("CARGO_PKG_VERSION"),);
 
 /// Create a new [`reqwest::Client`] with configuration appropriate to this library.
+#[cfg(not(target_arch = "wasm32"))]
 pub(super) fn client() -> Result<reqwest::Client> {
+    use std::time::Duration;
```
ryneeverett, line 21: I'm also curious if conditionally compiling `use` declarations actually results in smaller builds? Seems like the compiler would figure out that it isn't referenced.

ryneeverett (approved):

src/server/encryption.rs
djmitche: This was to get rid of an unused-item warning. I'm not sure what the best way to handle these would be -- I don't want to have a blanket `#![allow(unused)]`. Maybe just adding that attribute to each sometimes-used symbol would be better? At any rate, you're right about the simpler expression.

src/server/http.rs
djmitche: This was also moved to avoid an unused-import warning. It seemed better to move it closer to its usage than to add a conditional on the `use`.

I think (hope!) you're right about use declarations -- if they're unused, I'm sure they don't affect binary size.

src/server/encryption.rs
djmitche: Sorry, no, that isn't correct. We need this symbol if all of the server implementations are included. I'll leave it as-is for the moment. It's the sort of thing that's easy to change based on error messages if and when it needs to change.

PR #659, opened 2026-01-01
relax restrictions on tag chars after first, following TaskWarrior

smemsh:
This patch just allows the second and following chars for a tag to be anything except whitespace, which follows what TaskWarrior will be doing soon.

Restrictions remain on the first char.

A couple typos in the invalid char list are also fixed.

See https://github.com/GothenburgBitFactory/taskwarrior/pull/3961#pullrequestreview-3277625253


djmitche (approved):
Nice, thanks!

PR #661, opened 2026-01-05
Wait for wrapped storage to be dropped

### LINKED ISSUE #660: Send wrapper does not wait for storage to be dropped
The send wrapper closes its channel when it is dropped, which correctly causes the actor thread to wake up and drop the wrapped storage.

However, this is not synchronized: the send wrapper's drop completes without waiting for the actor thread to terminate. This means that if the caller exit()'s shortly after dropping the storage -- and Taskwarrior does just that -- then the storage never gets dropped.

This, in turn, means that the SqliteStorage never calls `sqlite_close`, which would flush WAL files and update the modification time on the `.sqlite3` file. Which means that even a command like `task add foo` might not update the modification time, which will probably mess with some users' workflows (it messes with mine!). It also causes some Taskwarrior tests to fail.

djmitche:
This removes the unnecessary `Clone` implementation from SqliteStorage, since JoinHandle is not clonable. I'm not sure why that was added. `SqliteStorage` did not implement `Clone` in 2.x, so I think we can "cheat" and remove this implementation in 3.0.1.

I've confirmed that this fixes the `test/read-only.test.py` tests in Taskwarrior.

### DIFF
--- a/src/storage/send_wrapper/wrapper.rs
+++ b/src/storage/send_wrapper/wrapper.rs
@@ -16,9 +16,12 @@ use uuid::Uuid;
 /// a few types are `!Send` and any async function handling such types are also `!Send`.
 ///
 /// On WASM, the wrapped storage runs in an async task, but not in a thread.
-#[derive(Clone)]
 pub(in crate::storage) struct Wrapper {
-    sender: mpsc::UnboundedSender<ActorMessage>,
+    // Both fields in this struct are `Option<..>` to allow them to be dropped individually
+    // in `Wrapper::drop`.
+    sender: Option<mpsc::UnboundedSender<ActorMessage>>,
+    #[cfg(not(target_arch = "wasm32"))]
+    thread: Option<std::thread::JoinHandle<()>>,
 }
 
 impl Wrapper {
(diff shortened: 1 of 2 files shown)


djmitche:
```
  type SqliteStorage is no longer UnwindSafe, in /home/runner/work/taskchampion/taskchampion/src/storage/sqlite/mod.rs:44
  type SqliteStorage is no longer RefUnwindSafe, in /home/runner/work/taskchampion/taskchampion/src/storage/sqlite/mod.rs:44
  type SqliteStorage is no longer UnwindSafe, in /home/runner/work/taskchampion/taskchampion/src/storage/sqlite/mod.rs:44
  type SqliteStorage is no longer RefUnwindSafe, in /home/runner/work/taskchampion/taskchampion/src/storage/sqlite/mod.rs:44
  type SqliteStorage no longer derives Clone, in /home/runner/work/taskchampion/taskchampion/src/storage/sqlite/mod.rs:44
```

I think all of these are OK to address in a bugfix release.  See [UnwindSafe](https://doc.rust-lang.org/core/panic/trait.UnwindSafe.html) for details on that auto-trait. I don't think users of this type are using it in `catch_unwind`.

ryneeverett (approved):

PR #690, opened 2026-02-24
Fixed edge case when a tag contains invalid characters

tmahmood:
A tag containing invalid character after first letter is considered a valid tag, instead of throwing an error. This PR resolves it

### DIFF
--- a/src/task/tag.rs
+++ b/src/task/tag.rs
@@ -70,7 +70,7 @@ impl FromStr for Tag {
         if !value
             .chars()
             .skip(1)
-            .all(|c| !(c.is_whitespace() || c == ':'))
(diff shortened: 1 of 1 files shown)


tmahmood:
Sorry, this is not applicable, so closing. I did not see TaskWarrior supports special characters in tags, but going through the documentation, it seems it supposed to. Even though it's not working in CLI. Probably, I am missing something.

djmitche:
This was updated recently in https://github.com/GothenburgBitFactory/taskchampion/pull/659 as part of https://github.com/GothenburgBitFactory/taskwarrior/pull/3961#pullrequestreview-3277625253. The latter PR has been merged but is not part of a release of TW yet.

tmahmood:
Got it, thank you

Issue #441, opened 2024-08-03
Sync to git


djmitche:
@lauft pointed out that the sync model would map well to the Git model. We could build a server implementation wrapping Git, using a similar set of files to those in the cloud storage: a `latest` file, a bunch of version files, and a bunch of snapshot files.

The `latest` file would be contentious: changing in every commit. However, that's precisely enforcing the replica invariant. Those conflicts can occur both on `git pull` (when new versions arrive) and on `git push` (if `latest` has changed since `git pull`).

djmitche:
This is probably a bit much for a "good first issue", but maybe consider it a "good second issue"!

djmitche:
I had a think about how we might want to do this. We need to implement the [Server trait](https://github.com/GothenburgBitFactory/taskchampion/blob/7143e2c5927de3aab4dd8770d6d4d360ec6b9910/taskchampion/src/server/types.rs), and specifically `add_version` and `get_child_version`.

The naming for versions, snapshots, and `latest` could match that of the GCP and AWS interfaces.

For `get_child_version`, a simple directory scan would suffice, with a `git pull` if the version is not found.

For `add_version`:
 * Check whether `latest` contains the `latest_version_id`; if not, `git pull` and check again. If still no, fail with `ExpectedParentVersion`
 * Prepare a new commit containing the new version and setting `latest` to the `latest_version_id`
 * `git push`. If this fails, revert the local commit, `git pull`, and fail with `ExpectedParentVersion` containing the new value in `latest`

I had a look at [git2](https://docs.rs/git2/latest/git2/), and it looks pretty complex even just to do these two operations. Maybe it's better to just shell out to `git` directly?

We'll also need to think about how this will be used, and abused. I suspect the folks requesting Git as a sync mechanism have very different ideas of how this might work, and would be tempted to muck about in the Git repo themselves.

The usage I'm aiming for is an "origin" repo, which might be GitHub or a repo on a VPS somewhere accessible from SSH. It is very much _not_ a peer-to-peer thing where changes are manually `git push` and `git pull`ed between repositories. In keeping with the other sync mechanisms, the version data would be stored in an encrypted form, so no other tools can read or manipulate it. `git diff` won't be any help. Replicas store locally the latest version_id they have seen in the repo, so using `git reset` to return to an earlier commit -- with an earlier version in `latest` -- will result in replicas being unable to sync.

[...]

Issue #685, opened 2026-02-13
Sync fails with IndexedDB due to weird Transaction semantics


Slug-Man:
When calling replica.sync() with IndexedDbStorage in a browser, sync fails with (in Chrome): `InvalidStateError: Failed to execute 'objectStore' on 'IDBTransaction': The transaction has finished.`

Root Cause:
The sync code in taskdb/sync.rs opens a single transaction and holds it open while:
1. Reading from storage (IDB operations)
2. Making HTTP calls to the sync server (network I/O)
3. Writing back to storage (IDB operations)

IndexedDB transactions auto-commit when there are no pending IDB operations. During HTTP calls, the browser sees no pending IDB work and closes the transaction. When sync tries to write afterward, it fails.

djmitche:
Thanks! I did miss that important detail in implementing this functionality, and it's a pretty fundamental flaw in the WASM support.

I suppose we have two options:
* Somehow make the `Txn` implementation satisfy the requirements of that trait -- perhaps by artificially holding the transaction open with repeated operations, or by layering on some kind of in-database locking like an `is_locked` table.
* Re-writing the sync algorithm so that it does not require the transaction to be open when contacting the remote server.

The first is impractical: busy-looping with repeated operations would waste a lot of time and energy, and a simple lock doesn't accomplish the not-visible-until-committed requirement.

The second might be possible. There are two remote calls involved in sync: [GetChildVersion](https://gothenburgbitfactory.org/taskchampion/sync-protocol.html#getchildversion) and [AddVersion](https://gothenburgbitfactory.org/taskchampion/sync-protocol.html#addversion). In both cases, we are already able to "crash" either before or after the server receives the request without data corruption. This implies that DB changes may occur while the API call proceeds: API call begins, process crashes, another process starts and makes changes. So, the questions is, do we rely on transaction rollback at any point -- that is, is there a point where something has been modified in the DB that must be rolled back if the response from the server is not as expected.

GetChildVersion is called having written nothing, so this is pretty easy. It depends on base_version, and could easily re-validate that value after the API call.

AddVersion takes a version, which we construct from the DB, but nothing is written to the DB before the API call. However, it does depend on the operations in that version not changing during the API call. We change operations by marking them as sync'd -- which would be OK, as it would result in duplicating operations on the server and this is already possible and handled properly.

In summary, it should be possible to re-write the sync process such that it works with shorter transactions that do not span over API calls.

djmitche:
[...]

PR #722, opened 2026-04-18
Add a git based sync option #441

carmiac:
This add a git backed sync server to TaskChampion.

## Repo Layout
- Versions are stored as files named `v-{parent_uuid}-{child_uuid}`, containing encrypted [`HistorySegment`] bytes.
- Snapshots are stored as a single file named `snapshot`, containing a JSON wrapper around an encrypted full-state blob.
- Metadata (`meta`) holds the latest version UUID and the encryption salt as JSON.

## Writes
After each write (`add_version`, `add_snapshot`) the server stages the changed files, creates a commit, and pushes to the remote. If the push is rejected , the commit is rolled back and the caller receives an [`AddVersionResult::ExpectedParentVersion`] or an [`Error`] so it can retry.

After a snapshot is stored, [`GitSyncServer::cleanup`] automatically removes all version files whose history is now captured by the snapshot, keeping the repository compact.

## Options

There are several configuration options.

 - `branch` sets the branch to use for TaskChampion. This would let someone keep tasks alongside a project without dealing with history issues.
 - `remote` sets the remote. Same format the git uses.
 - `local_only` makes `push` and `pull` no-ops. This could be used if the remote isn't ready yet, or is unavailable for some other reason.

## General Notes for Reviewers
 - I haven't done any performance testing, but it seems reasonably quick for manual use. I didn't see a general performance testing module.
 - Currently is uses the same salt for all files. This isn't great security practice, but does seem to be what the other servers are doing. I could add a per-file salt, at the cost of however long Cryptor::new() takes on each read and write.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -49,7 +49,7 @@ server-aws = [
 # Suppport for sync to another SQLite database on the same machine
 server-local = ["dep:rusqlite"]
 # Support for sync via Git
-server-git = ["git-sync"]
+server-git = ["dep:serde_with", "dep:glob", "encryption"]
 
 # Support for all task storage backends (except indexeddb, which only works on WASM builds)
 storage = ["storage-sqlite"]
@@ -62,8 +62,6 @@ storage-indexeddb = ["dep:idb", "dep:web-sys"]
 encryption = ["dep:ring"]
 # (private) Generic support for cloud sync
 cloud = []
-# (private) Support for Git based sync
-git-sync = ["dep:serde_with", "dep:glob", "encryption"]
 # static bundling of dependencies
 bundled = ["rusqlite/bundled"]
 # use CA roots built into the library for all HTTPS access
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -63,7 +63,7 @@ encryption = ["dep:ring"]
 # (private) Generic support for cloud sync
 cloud = []
 # (private) Support for Git based sync
-git-sync = ["dep:hex"]
+git-sync = ["dep:serde_with", "dep:glob"]
 # static bundling of dependencies
 bundled = ["rusqlite/bundled"]
 # use CA roots built into the library for all HTTPS access
@@ -83,7 +83,7 @@ async-trait = "0.1.89"
 byteorder = "1.5"
 chrono = { version = "^0.4.44", features = ["serde"] }
 flate2 = "1"
-hex = { version = "0.4", optional = true }
+serde_with = { version = "3.18.0", features = ["base64"], optional = true }
 idb = { version = "0.6.4", optional = true }
 log = "^0.4.17"
 # Reqwest is "stuck" at 0.12 because that's what google-cloud-storage depends on.
@@ -98,7 +98,7 @@ tokio = { version = "1", features = ["macros", "sync", "rt"] }
 thiserror = "2.0"
 uuid = { version = "^1.23.0", features = ["serde", "v4"] }
 url = { version = "2", optional = true }
-glob = "0.3.3"
+glob = { version = "0.3.3", optional = true }
 
 ## wasm-only dependencies.
 [target.'cfg(target_arch = "wasm32")'.dependencies]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -98,6 +98,7 @@ tokio = { version = "1", features = ["macros", "sync", "rt"] }
 thiserror = "2.0"
 uuid = { version = "^1.23.0", features = ["serde", "v4"] }
 url = { version = "2", optional = true }
+glob = "0.3.3"
 
 ## wasm-only dependencies.
 [target.'cfg(target_arch = "wasm32")'.dependencies]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -128,7 +128,7 @@ web-sys = { version = "0.3.83", optional = true, features = [
 
 ## non-wasm dependencies
 [target.'cfg(not(target_arch = "wasm32"))'.dependencies]
-aws-sdk-s3 = { version = "1.127", default-features = false, optional = true }
+aws-sdk-s3 = { version = "1.128", default-features = false, optional = true }
 aws-config = { version = "1.8", default-features = false, features = [
   "rt-tokio",
   "behavior-version-latest",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,7 +14,7 @@ edition = "2021"
 rust-version = "1.91.1"
 
 [workspace]
-members = [ "xtask" ]
+members = ["xtask"]
 resolver = "2"
 
 [lib]
@@ -24,15 +24,32 @@ crate-type = ["cdylib", "rlib"]
 default = ["sync", "bundled", "storage", "tls-webpki-roots"]
 
 # Support for all sync solutions
-sync = ["server-sync", "server-gcp", "server-aws", "server-local"]
+sync = ["server-sync", "server-gcp", "server-aws", "server-local", "server-git"]
 # Support for sync to a server
 server-sync = ["encryption", "http"]
 # Support for sync to GCP
-server-gcp = ["cloud", "encryption", "http", "dep:google-cloud-storage", "dep:reqwest-middleware"]
+server-gcp = [
+  "cloud",
+  "encryption",
+  "http",
+  "dep:google-cloud-storage",
+  "dep:reqwest-middleware",
+]
 # Support for sync to AWS
-server-aws = ["cloud", "encryption", "http", "dep:aws-sdk-s3", "dep:aws-config", "dep:aws-credential-types", "dep:aws-smithy-runtime-api", "dep:aws-smithy-types"]
+server-aws = [
+  "cloud",
+  "encryption",
+  "http",
+  "dep:aws-sdk-s3",
+  "dep:aws-config",
+  "dep:aws-credential-types",
+  "dep:aws-smithy-runtime-api",
+  "dep:aws-smithy-types",
+]
 # Suppport for sync to another SQLite database on the same machine
 server-local = ["dep:rusqlite"]
+# Support for sync via Git
+server-git = ["git-sync"]
 
 # Support for all task storage backends (except indexeddb, which only works on WASM builds)
 storage = ["storage-sqlite"]
@@ -45,6 +62,8 @@ storage-indexeddb = ["dep:idb", "dep:web-sys"]
 encryption = ["dep:ring"]
 # (private) Generic support for cloud sync
 cloud = []
+# (private) Support for Git based sync
+git-sync = ["dep:hex"]
 # static bundling of dependencies
 bundled = ["rusqlite/bundled"]
 # use CA roots built into the library for all HTTPS access
@@ -64,12 +83,13 @@ async-trait = "0.1.89"
 byteorder = "1.5"
 chrono = { version = "^0.4.44", features = ["serde"] }
 flate2 = "1"
+hex = { version = "0.4", optional = true }
 idb = { version = "0.6.4", optional = true }
 log = "^0.4.17"
 # Reqwest is "stuck" at 0.12 because that's what google-cloud-storage depends on.
 reqwest = { version = "0.12", default-features = false, optional = true }
 ring = { version = "0.17", optional = true }
-rusqlite = { version = "0.39", features = [ "fallible_uint" ], optional = true}
+rusqlite = { version = "0.39", features = ["fallible_uint"], optional = true }
 serde_json = "^1.0"
 serde = { version = "^1.0.147", features = ["derive"] }
 strum = "0.28"
@@ -82,7 +102,9 @@ url = { version = "2", optional = true }
 ## wasm-only dependencies.
 [target.'cfg(target_arch = "wasm32")'.dependencies]
 uuid = { version = "1", features = ["js"] }
-ring = { version = "0", optional = true, features = ["wasm32_unknown_unknown_js"] }
+ring = { version = "0", optional = true, features = [
+  "wasm32_unknown_unknown_js",
+] }
 getrandom = { version = "0.4", features = ["wasm_js"] }
 wasm-bindgen = { version = "0.2" }
 wasm-bindgen-futures = { version = "0.4" }
@@ -101,16 +123,24 @@ web-sys = { version = "0.3.83", optional = true, features = [
   "EventTarget",
   "Event",
   "console",
-]}
+] }
 
 ## non-wasm dependencies
 [target.'cfg(not(target_arch = "wasm32"))'.dependencies]
 aws-sdk-s3 = { version = "1.127", default-features = false, optional = true }
-aws-config = { version = "1.8", default-features = false, features = ["rt-tokio", "behavior-version-latest"], optional = true }
-aws-credential-types = { version = "1", default-features = false, features = ["hardcoded-credentials"], optional = true }
+aws-config = { version = "1.8", default-features = false, features = [
+  "rt-tokio",
+  "behavior-version-latest",
+], optional = true }
+aws-credential-types = { version = "1", default-features = false, features = [
+  "hardcoded-credentials",
+], optional = true }
 aws-smithy-runtime-api = { version = "1.11", default-features = false, optional = true }
 aws-smithy-types = { version = "1.4", default-features = false, optional = true }
-google-cloud-storage = { version = "0.24.0", default-features = false, features = ["rustls-tls", "auth"], optional = true }
+google-cloud-storage = { version = "0.24.0", default-features = false, features = [
+  "rustls-tls",
+  "auth",
+], optional = true }
 reqwest-middleware = { version = "0.4", optional = true }
 
 ## common dev-dependencies
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -63,7 +63,7 @@ encryption = ["dep:ring"]
 # (private) Generic support for cloud sync
 cloud = []
 # (private) Support for Git based sync
-git-sync = ["dep:serde_with", "dep:glob"]
+git-sync = ["dep:serde_with", "dep:glob", "encryption"]
 # static bundling of dependencies
 bundled = ["rusqlite/bundled"]
 # use CA roots built into the library for all HTTPS access
--- a/src/server/gitsync/mod.rs
+++ b/src/server/gitsync/mod.rs
@@ -2,31 +2,51 @@ use crate::errors::Result;
 use crate::server::encryption::{Cryptor, Sealed, Unsealed};
 use crate::server::{
     AddVersionResult, GetVersionResult, HistorySegment, Server, Snapshot, SnapshotUrgency,
-    VersionId, NIL_VERSION_ID,
+    VersionId,
 };
 use crate::Error;
 use async_trait::async_trait;
 use glob::glob;
 use log::info;
 use serde::{Deserialize, Serialize};
+use serde_with::{base64::Base64, serde_as};
 use std::fs::{self, File};
-use std::io::Read;
+use std::io::{BufReader, Read};
 use std::path::{Path, PathBuf};
 use std::process::Command;
+use std::time::TryFromFloatSecsError;
 use uuid::Uuid;
+
+#[serde_as]
 #[derive(Serialize, Deserialize, Debug)]
 struct Version {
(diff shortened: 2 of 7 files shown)


djmitche (commented):
This is so cool! I would like to take the time to read and experiment with it, but a few initial reactions to the comments and the PR description:

 * Optional configuration for the path to the `git` binary will probably be useful for someone, and should be easy to add
 * It looks like `remote` accepts the string "None"? Let's make that an `Option<String>` instead, if that's the case, or make the empty string the special case.
 * For cleanup, two observations:
    * Git keeps all of the data anyway, so "deleting" a version is really just removing the name from the directory. So, I don't think deleting old versions has much impact.
    * When a replica syncs, it needs all versions since the last version it saw. Otherwise, it has to restore from the snapshot and lose any local data since it last sync'd.
    * So, it's beneficial to keep versions around for as long as is practical. The cloud servers keep a half-year's worth of versions (`src/server/cloud/server.rs`, `MAX_VERSION_AGE_SECS`), and I think that's reasonable here, too.

I'd be interested to know what @ryneeverett thinks, too!

carmiac:
Thanks for the review! I'll take a look at those. My reason behind the git rm was to make the globs when adding new versions faster, but I didn't realize how that interacted with restoration.

djmitche:
Hm, good point, but local globs are probably pretty fast even for 100's of dentries.

Cargo.toml
```diff
@@ -24,15 +24,32 @@ crate-type = ["cdylib", "rlib"]
...
+  "dep:aws-sdk-s3",
+  "dep:aws-config",
+  "dep:aws-credential-types",
+  "dep:aws-smithy-runtime-api",
+  "dep:aws-smithy-types",
+]
 # Suppport for sync to another SQLite database on the same machine
 server-local = ["dep:rusqlite"]
+# Support for sync via Git
+server-git = ["git-sync"]
```
djmitche, line 52: Why the two levels of features here? I think one would be sufficient. The cloud syncs have `cloud` because it enables some common functionality, but there's no such thing for git. So,

```
server-git = ["dep:serde_with", "dep:glob", "encryption"]
```
and update the `cfg(feature..)` in the code.

src/server/gitsync/mod.rs
```diff
@@ -0,0 +1,1149 @@
...
+//! - This does support both defining a remote and having `local_only` mode set at the same
+//!   time. The idea is that maybe the remote isn't ready yet, or eithe rtemporarily or
+//!   permanantly down. Either way, you can use this in local mode in the mean time.
+//! - Remember, a remote can be on the same machine as local. This is used for testing.
+//!
+//! Notes for Reviewers
+//!
+//! - I haven't done any performance testing, but it seems reasonably quick for manual use.
+//! - Currently is uses the same salt for all files. This isn't great security practice,
+//!   but does seem to be what the other servers are doing.
```
djmitche, line 40: This is a good point! In defense of the idea, the salt is used in the key derivation, and we use the same key for every file, so in that sense it's only used once. If there's further concern, let's open an issue about it -- I'm sure others would like to chime in too!

src/server/gitsync/mod.rs
```diff
@@ -0,0 +1,1149 @@
...
+//! - If you are planning on using it for other things, it is HIGHLY recommended that you
+//!   create a 'task' branch and let TaskChampion manage that branch.
+//! - This does support both defining a remote and having `local_only` mode set at the same
+//!   time. The idea is that maybe the remote isn't ready yet, or eithe rtemporarily or
+//!   permanantly down. Either way, you can use this in local mode in the mean time.
+//! - Remember, a remote can be on the same machine as local. This is used for testing.
+//!
+//! Notes for Reviewers
+//!
+//! - I haven't done any performance testing, but it seems reasonably quick for manual use.
```
djmitche, line 38: This seems fine. We have not focused on performance testing of the sync operations since most of the time is network-related.

src/server/gitsync/mod.rs
```diff
@@ -0,0 +1,1149 @@
...
+//!   setup. I.e. 'git init', 'git add', 'git commit', etc shoud just work.
+//! - If you are using a remote, 'git push' and 'git pull' shoud work.
+//! - Due to the nature of the version and snapshot history, you probably shouldn't do
+//!   a lot of the things you normally would with a git repo, like merge, squash, etc.
+//!   Just let TaskChampion manage it.
+//! - If you are planning on using it for other things, it is HIGHLY recommended that you
+//!   create a 'task' branch and let TaskChampion manage that branch.
+//! - This does support both defining a remote and having `local_only` mode set at the same
+//!   time. The idea is that maybe the remote isn't ready yet, or eithe rtemporarily or
+//!   permanantly down. Either way, you can use this in local mode in the mean time.
```
djmitche, line 33: What are the risks here? In general, the more user-facing bits of the doc here might be better as docstrings on the configuration enum.

src/server/gitsync/mod.rs
```diff
@@ -0,0 +1,1149 @@
...
+//!
+//! After a snapshot is stored, [`GitSyncServer::cleanup`] automatically removes all
+//! version files whose history is now captured by the snapshot, keeping the repository
+//! compact.
+//!
+//! Notes and Expectations
+//!
+//! - Since this shells out to git, it assumes that you havea reasonably functional git
+//!   setup. I.e. 'git init', 'git add', 'git commit', etc shoud just work.
+//! - If you are using a remote, 'git push' and 'git pull' shoud work.
```
djmitche, line 25: I suspect there's some room for more robustness here, such as disabling prompts. Maybe we can address that as we find issues, but it might be worth thinking about ahead of time.

I see that nothing needs to parse the output of a `git` command, so that simplifies things!

src/server/gitsync/mod.rs
```diff
@@ -0,0 +1,1149 @@
...
+        let meta_path = self.write_meta()?;
+
+        // Commit and push, reverting if push fails.
+        self.stage_and_commit(&[&version_path, &meta_path], "add version")?;
+
+        if !self.push()? {
+            // Push was rejected. Undo the commit; pull will reset --hard and git clean
+            // away the stray version file.
+            git_cmd(&self.local_path, &["reset", "HEAD~1", "--soft"])?;
+            self.pull()?;
```
djmitche, line 562: If this repo had been in `local_only` mode for a while, this pull might not be fast-forwardable, in which case we have two Git histories with two sync histories, which are basically irreconcilable. Basically, if we ever return from `add_version` with a commit in the local repo that is not also in the remote repo, we run this risk: we've promised the local replica that the Version is published to the sync server, but have not actually done so.

This seems like a hidden risk for `local_only`. Perhaps docs for that parameter should suggest using it only for setup, or otherwise explain this risk?

src/server/gitsync/mod.rs
```diff
@@ -0,0 +1,1149 @@
...
+fn parse_version_filename(name: &str) -> Option<(Uuid, Uuid)> {
+    let stem = name.strip_prefix("v-")?;
+    let (parent_str, child_str) = stem.split_once('-')?;
+    let parent_id = Uuid::parse_str(parent_str).ok()?;
+    let child_id = Uuid::parse_str(child_str).ok()?;
+    Some((parent_id, child_id))
+}
+
+/// Run a git command in a given directory, returning an error if it exits non-zero.
+fn git_cmd(dir: &Path, args: &[&str]) -> Result<()> {
```
djmitche, line 122: This ends up putting a lot of `git`'s output in the `cargo test` output and probably in the `taskwarrior` output as well. What do you think of amending this function so that it only shows the output on an unexpected error, or logs it all to `log::debug` or something similar?

src/server/gitsync/mod.rs
```diff
@@ -0,0 +1,1149 @@
...
+    /// Initialise or open the git repository and return the current [`Meta`].
+    ///
+    /// Creates the directory, initialises or clones the repo, switches to `branch`, and
+    /// creates the `meta` file on first run.
+    fn init_repo(local_path: &Path, branch: &str, remote: &str, local_only: bool) -> Result<Meta> {
+        // Create the local directory if needed.
+        fs::create_dir_all(local_path)?;
+
+        // Check if path is already a git repo.
+        let is_repo = Command::new("git")
```
djmitche, line 170: I think this doesn't use `git_cmd` because a nonzero exit status is expected, and similar for `git checkout` below. Maybe `git_cmd` could be extended to support that situation?

src/server/gitsync/mod.rs
```diff
@@ -0,0 +1,1149 @@
...
+                git_cmd(local_path, &["commit", "-m", "init taskchampion repo"])?;
+                m
+            }
+        };
+
+        // Remove any untracked files left behind by interrupted writes.
+        git_cmd(
+            local_path,
+            &["clean", "-f", "--", "v-*", "snapshot", "meta"],
+        )?;
```
djmitche, line 248: This happens at least twice and could be a helper function!

src/server/gitsync/mod.rs
```diff
@@ -0,0 +1,1149 @@
...
+    fn write_meta(&self) -> Result<PathBuf> {
+        let meta_path = self.local_path.join("meta");
+        let f = File::create(&meta_path)?;
+        serde_json::to_writer(f, &self.meta)?;
+        Ok(meta_path)
+    }
+
+    /// Fetch and fast-forward to the remote branch. No-op in local-only mode.
+    /// If the remote branch does not yet exist (e.g. fresh bare repo), this is also a no-op.
+    fn pull(&self) -> Result<()> {
```
djmitche, line 281: This isn't really a pull, since it does a hard reset. Maybe `fn reset_to_remote`?

src/server/gitsync/mod.rs
```diff
@@ -0,0 +1,1149 @@
...
+        }
+
+        // Create the new version and write it to file.
+        let version_id = Uuid::new_v4();
+        let version = Version {
+            version_id,
+            parent_version_id,
+            history_segment,
+        };
+        let version_path = self.add_version_by_parent_version_id(&version)?;
```
djmitche, line 551: For robustness, should this check that the working copy is clean before starting modifications? If a previous invocation of `add_version` crashed right here, we'd end up with two different version files being committed.

djmitche (changes requested):
A bunch of comments here, but all toward improving the implementation rather than show-stopping issues. I think most of what I've suggested is relatively straightforward, but if necessary some of it can be done in followup PRs.

One general observation is, despite Server being an async trait, this invokes Git synchronously. I think that's fine for the expected use-cases for this sync model, and it's something that can be improved later if desired.

tl;dr: This looks great, and I look forward to merging it after some minor revisions!

src/server/gitsync/mod.rs
```diff
@@ -0,0 +1,1149 @@
...
+//! rolled back and the caller receives an [`AddVersionResult::ExpectedParentVersion`]
+//! or an [`Error`] so it can retry.
+//!
+//! After a snapshot is stored, [`GitSyncServer::cleanup`] automatically removes all
+//! version files whose history is now captured by the snapshot, keeping the repository
+//! compact.
+//!
+//! Notes and Expectations
+//!
+//! - Since this shells out to git, it assumes that you havea reasonably functional git
```
ryneeverett, line 23:
```suggestion
//! - Since this shells out to git, it assumes that you have a reasonably functional git
```

src/server/gitsync/mod.rs
```diff
@@ -0,0 +1,1149 @@
...
+//!   setup. I.e. 'git init', 'git add', 'git commit', etc shoud just work.
+//! - If you are using a remote, 'git push' and 'git pull' shoud work.
+//! - Due to the nature of the version and snapshot history, you probably shouldn't do
+//!   a lot of the things you normally would with a git repo, like merge, squash, etc.
+//!   Just let TaskChampion manage it.
+//! - If you are planning on using it for other things, it is HIGHLY recommended that you
+//!   create a 'task' branch and let TaskChampion manage that branch.
+//! - This does support both defining a remote and having `local_only` mode set at the same
+//!   time. The idea is that maybe the remote isn't ready yet, or eithe rtemporarily or
```
ryneeverett, line 32:
```suggestion
//!   time. The idea is that maybe the remote isn't ready yet, or either temporarily or
```

src/server/gitsync/mod.rs
```diff
@@ -0,0 +1,1149 @@
...
+        // only create a new one if that fails.
+        info!("Switching branch to {:?}", branch);
+        let checkout_ok = Command::new("git")
+            .args(["checkout", branch])
+            .current_dir(local_path)
+            .stderr(std::process::Stdio::null())
+            .status()?
+            .success();
+        if !checkout_ok {
+            // For a brand-new repo `git checkout -b` also fails, so use
```
ryneeverett, line 209: By "brand-new repo" do you more specifically mean "bare repo with no commits"? I presume you realize that `git init && git checkout -b <branch>` doesn't fail.

src/server/gitsync/mod.rs
```diff
@@ -0,0 +1,1149 @@
...
+            Ok(m) => m,
+            Err(_) => {
+                let m = Meta {
+                    latest_version: Uuid::nil(),
+                    salt: Cryptor::gen_salt()?,
+                };
+                let f = File::create_new(&meta_path)?;
+                serde_json::to_writer(f, &m)?;
+                git_cmd(local_path, &["add", "meta"])?;
+                git_cmd(local_path, &["commit", "-m", "init taskchampion repo"])?;
```
ryneeverett, line 239: Maybe use `stage_and_commit` here?

ryneeverett (commented):
I don't see any mention of purging old versions or snapshots which suggests they will live on in git history indefinitely. This seems somewhat worse than other implementations which may not purge snapshots but at least truly delete versions. This doesn't necessarily need to be addressed in the initial implementation unless it informs the design. Have we thought about this?

src/server/gitsync/mod.rs
```diff
@@ -206,7 +206,7 @@ impl GitSyncServer {
             .status()?
             .success();
         if !checkout_ok {
-            // For a brand-new repo `git checkout -b` also fails, so use
+            // For a repo with no commits yet, `git checkout -b` also fails, so use
```
ryneeverett: This doesn't clarify it -- at least not for me. Again, `git init && git checkout -b foo` does not fail. You specifically mean a **bare** repo, right?

ryneeverett (commented):
My inline comments were all addressed, thanks!

> I don't see any mention of purging old versions or snapshots which suggests they will live on in git history indefinitely. This seems somewhat worse than other implementations which may not purge snapshots but at least truly delete versions. This doesn't necessarily need to be addressed in the initial implementation unless it informs the design. Have we thought about this?

Thinking about this a second time, I don't think this is relevant to the design. One way or another it involves deleting commit history and the repository structure is irrelevant to that.

src/server/gitsync/mod.rs
```diff
@@ -0,0 +1,1341 @@
...
+                .to_str()
+                .ok_or_else(|| Error::Server("path is not valid UTF-8".into()))?;
+            self.cmd(dir, &["add", path_str])?;
+        }
+        self.cmd(dir, &["commit", "-m", message])
+    }
+}
+
+/// Remove untracked TaskChampion files left behind by interrupted writes.
+fn clean_stray_files(git: &Git, dir: &Path) -> Result<()> {
```
djmitche: Is there a rationale for `stage_and_commit` being a method while this is a module-level function?

djmitche (approved):
This looks good. Can you do a quick rebase or merge? One comment below but it's not blocking merging.

ryneeverett:
Sorry but I can't stop thinking about the disk size issue. Maybe snapshots should be disabled entirely for the git backend? Whereas they (eventually) save disk space on other backends, on git they only serve to increase the disk space and data transfer.

src/server/gitsync/mod.rs
carmiac: Nope, I just missed it.

carmiac:
> Sorry but I can't stop thinking about the disk size issue. Maybe snapshots should be disabled entirely for the git backend? Whereas they (eventually) save disk space on other backends, on git they only serve to increase the disk space and data transfer.

I think there are several options for handling disk size with a git backend.
1) What it does now. Disk size increases unbounded on all replicas, but the implementation is relatively simple and doesn't cause a large slowdown for new replicas of existing servers.
2) No snapshots. Disk size still increases unbounded, but at a slower rate. Implementation is even easier but new replicas of existing servers are slow to initialize.
3) Shallow clones only. Local disk space is more or less bounded, but origin still grows unbounded. Will take a bit of thinking about how to handle local only mode.
4) Purge old files via 'git filter-repo'. All repos have a more-or-less bounded size, but push/pull starts to get dicey as the repo history is rewritten.
5) Something tricky that combines 3 and 4, with only origin purging. It will take some thinking that I'm not capable of right now due to having a cold.
6) Something tricky with re-initializing origin every so often. Also not capable of figuring that our right now.

ryneeverett:
> 2. No snapshots. Disk size still increases unbounded, but at a slower rate. Implementation is even easier but new replicas of existing servers are slow to initialize.

How confident are we that new replicas are slower to initialize? The assumption is that it is faster to download and index a snapshot than to traverse the additional versions it squashed, right?

carmiac:
Completely untested by me. I'm only basing it on the comments in the other servers code and the TaskChampion book. But it does make sense that having to walk the entire history will be slower than starting from a reasonably recent point and only walking through updates since then. Though with also having to pull a full git repo that may not be true.

djmitche:
I think we can push this particular problem into the future, rather than solving it here -- and I think keeping snapshots is the most future-compatible way to do that. Let's wrap this PR up and open an issue for bounding size.

ryneeverett:
If bounding size is actually going to happen then I agree that snapshots are essential. But if the problem is being pushed into the indefinite (possibly never) future then it might be better to reduce the impact of unbound size.

djmitche:
That's fair, but I'll be optimistic (which is unusual for me) and say it's better to plan for the future being positive than to assume it will not. And, I worry that something like not including snapshots would leave us with library users that _depend_ on the full history, meaning however well-intentioned future devs are, they can't fix this without breaking things.

Realistically, this seems like like something bite-sized that any of the three of us, given a bit of time, could do -- as could any motivated contributor. So if we see users suffering from excessive repo sizes, I think we'll see some pressure to fix it and that the fix will happen.

Now, [the WASM sync issues](https://github.com/GothenburgBitFactory/taskchampion/issues/685) are another matter entirely...

ryneeverett:
If it is that easy then great. I haven't come up with any better ideas than the ones @carmiac listed though and it isn't clear that there is a robust option. I guess worst case scenario is the client rebases and force pushes the working tree as the initial commit?

I was thinking that unbounded size might actually be tolerable for the vast majority of users who would choose git. I would see this as a alternative positive outcome.

carmiac:
I will say that for my use case, unbounded repo size doesn't really matter. If it ever gets too ridiculous, I could manually clean up the remote and then re-init the locals. It would be interesting to see what the average storage size per 1000 tasks is to help get a handle on how much bike shedding is worth it. 

FWIW, in 2020 torvalds/linux took up about 4GB on disk with over a million commits. I think we'll be fine.

ryneeverett:
Frankly, I think your position supports my point -- we'll probably never be motivated to address this issue. So does it make sense to have the repo size grow many times faster in order to make it easier to stop the growth "someday maybe"?

djmitche:
I can see a path here where we don't _store_ snapshots, but still look for them when initializing a replica. Then if we decide to do the followup work, we can start creating snapshots and ensure there is an active snapshot before deleting old data.

However, that seems higher-risk than the alternative of landing this and following up (either soon or when there's user demand) with a mechanism for removing stale data.

In particular, I don't think we have enough data to understand the performance of applying many, many versions when initializing a replica. Nor do we have enough data to be confident that storing versions but not snapshots is likely to provide enough space savings to make this a non-issue. Taskwarrior users have surprisingly large and frequently-changing task DB's.

I'm inclined to land this PR more-or-less as-is, mark the issue for later followup, and see what happens.

carmiac:
Is there any more work needed for this?

djmitche:
Just rebase or merge to avoid the conflicts and I'm happy to merge to main.

carmiac:
Done!

djmitche:
Woo, we have a Git backend!

Do you mind opening issues for the followups?

PR #757, opened 2026-07-24
Don't build a dep map (twice!) when creating a new task.

carmiac:
Bulk creating tasks is O(n^2) because of a redundant depmap scan.  

`Replica::create_task` builds the full depmap for a brand new task, but that map is only ever queried for the tasks own UUID and a brand new task has no dependencies or dependents. 

The map is invalidated on every `commit_operations`, so bulk creating tasks one at a time (like `TDB2::add` does) does a full scan on every task. This is the cause of the test timeout on https://github.com/GothenburgBitFactory/taskwarrior/pull/4157

This changes the new task creation to skip the scan and use an empty depmap, which will have the same results for a new task without the scan. Existing tasks still get the real depmap.

### DIFF
--- a/src/replica.rs
+++ b/src/replica.rs
@@ -308,11 +308,17 @@ impl<S: Storage> Replica<S> {
     /// Use [Uuid::new_v4] to invent a new task ID, if necessary. If the task already
     /// exists, it is returned.
     pub async fn create_task(&mut self, uuid: Uuid, ops: &mut Operations) -> Result<Task> {
-        if let Some(task) = self.get_task(uuid).await? {
-            return Ok(task);
+        // Check for task existence without building the depmap if it doen't exist.
+        if let Some(taskdata) = self.get_task_data(uuid).await? {
+            let depmap = self.dependency_map(false).await?;
+            return Ok(Task::new(taskdata, depmap));
         }
-        let depmap = self.dependency_map(false).await?;
-        Ok(Task::new(TaskData::create(uuid, ops), depmap))
+        // A brand new task doesn't have dependencies or dependants, so creating an empty
+        // map is fine and much faster.
+        Ok(Task::new(
+            TaskData::create(uuid, ops),
+            Arc::new(DependencyMap::new()),
+        ))
     }
 
     /// Create a new, empty task with the given UUID.  This is useful for importing tasks, but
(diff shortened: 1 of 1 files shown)


src/replica.rs
```diff
@@ -308,11 +308,17 @@ impl<S: Storage> Replica<S> {
     /// Use [Uuid::new_v4] to invent a new task ID, if necessary. If the task already
     /// exists, it is returned.
     pub async fn create_task(&mut self, uuid: Uuid, ops: &mut Operations) -> Result<Task> {
-        if let Some(task) = self.get_task(uuid).await? {
-            return Ok(task);
+        // Check for task existence without building the depmap if it doen't exist.
+        if let Some(taskdata) = self.get_task_data(uuid).await? {
+            let depmap = self.dependency_map(false).await?;
```
djmitche, line 313: The `dependency_map` function does nothing if the Replica already has a valid dependency map, so calling it twice isn't actually an issue.

src/replica.rs
```diff
@@ -308,11 +308,17 @@ impl<S: Storage> Replica<S> {
...
+            return Ok(Task::new(taskdata, depmap));
         }
-        let depmap = self.dependency_map(false).await?;
-        Ok(Task::new(TaskData::create(uuid, ops), depmap))
+        // A brand new task doesn't have dependencies or dependants, so creating an empty
+        // map is fine and much faster.
+        Ok(Task::new(
+            TaskData::create(uuid, ops),
+            Arc::new(DependencyMap::new()),
```
djmitche, line 321: This is incorrect, although in a relatively harmless fashion. The depmap is a map for _all_ tasks, although it's currently only used for dependencies related to this task, of which there are none in htis case.

djmitche (commented):
I'm not convinced this is actually fixing a problem. Is the `dependency_map` function actually making the same DB queries more than once, or is it just called more than once? If it's making the DB queries more than once, then is something invalidating it between those fetches? The invalidation occurs in the `commit_operations` and `commit_reversed_operations` methods.

src/replica.rs
carmiac: Ah, I missed the caching semantics, thanks!

src/replica.rs
carmiac: Good point, I've updated the comment to better explain the change.

carmiac:
> Is the `dependency_map` function actually making the same DB queries more than once, or is it just called more than once? If it's making the DB queries more than once, then is something invalidating it between those fetches? The invalidation occurs in the `commit_operations` and `commit_reversed_operations` methods.

Yes, the invalidation is called in TaskWarriors `TDB2::add` when using `Task` rather than raw `TaskData`, which makes bulk task creation slow. I've updated the PR description and comments with the full logic.

djmitche (approved):
Thanks for the updates and helping me understand the situation :)

PR #762, opened 2026-08-15
Fix sqlite on haiku

carmiac:
With these changes it passes all tests on a recent Haiku nightly. I will add a CI test in a separate PR once the beta6 github action exists.

- WAL mode is broken on Haiku due to a posix bug
- Don't immediately drop the tmp_dir before it is used in the test helper

### DIFF
--- a/src/storage/sqlite/mod.rs
+++ b/src/storage/sqlite/mod.rs
@@ -76,8 +76,8 @@ mod test {
     use tempfile::TempDir;
 
     async fn storage() -> Result<SqliteStorage> {
-        let tmp_dir = TempDir::new()?;
-        SqliteStorage::new(tmp_dir.path(), AccessMode::ReadWrite, true).await
+        let tmp_dir = TempDir::new()?.keep();
+        SqliteStorage::new(tmp_dir.as_path(), AccessMode::ReadWrite, true).await
     }
 
     crate::storage::test::storage_tests!(storage().await?);
--- a/src/storage/sqlite/mod.rs
+++ b/src/storage/sqlite/mod.rs
@@ -72,12 +72,14 @@ mod test {
     use super::*;
     use crate::errors::Error;
     use crate::storage::config::AccessMode;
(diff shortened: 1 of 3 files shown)


src/storage/sqlite/mod.rs
```diff
@@ -76,8 +76,8 @@ mod test {
     use tempfile::TempDir;
 
     async fn storage() -> Result<SqliteStorage> {
-        let tmp_dir = TempDir::new()?;
```
djmitche, line 79: > This consumes the [TempDir](https://docs.rs/tempfile/latest/tempfile/struct.TempDir.html) without deleting directory on the filesystem, meaning that the directory will no longer be automatically deleted.

This automatic cleanup means that devs' disks don't fill up with old sqlite DBs.

As written, the temp dir is created, SQLite starts and opens its files in that dir, and then the dir is deleted (leaving the open files still accessible via their fd's). What fails on Haiku?

We may need a larger refactor to scope the TempDir around the entire test execution, or scan for and delete old temp dirs on every test run. Both would be pretty annoying and error-prone!

Maybe another option is to make this leaking-temp-dirs behavior conditional on `target_os = "haiku"`, so other developers continue to get test runs cleaning up after themselves?

djmitche (changes requested):
Other than the note below, this looks good!

src/storage/sqlite/mod.rs
carmiac: I've added a wrapper that makes sure the tmp_dir gets cleaned up.

The issue is that in WAL mode, SQLite doesn't notice that it's directory is being deleted, but in-non WAL mode it does. I think this is because in WAL mode it's keeping the directory open so it remains until the open reference is gone, but in non-WAL mode it's not.

djmitche: That's clever!!

djmitche (approved):
Other than the note below, this looks good!

